PR #7, opened 2021-07-19
feat: client-side exposure tracking

bgiori:
### Summary

First pass implementation of client-side exposure tracking

### Checklist


packages/browser/src/types/provider.ts
```diff
@@ -0,0 +1,20 @@
...
+  getUser(): ExperimentUser;
+}
+
+/**
+ * Provides a tracking implementation for standard experiment events generated
+ * by the client (e.g. exposure).
+ * @category Provider
+ */
+export interface ExperimentTrackingProvider {
+  track(eventName: string, properties: Record<string, string>): void;
```
ManThursday, line 19: I wondering about:
1) `track(eventName, properties)` vs `track(key, variant)`
2) `track(eventName, ...)` vs `trackExposure(...)`

My thinking about the alternative from (1) is that it it doesn't assume anything about how an alternative provider might be implemented.
All of the Amplitude instrumentation specific logic can be contained in the default provider and customer created providers can either copy that or directly use the default provider.

(2) is partly about enabling the API change from (1) and also just slightly different aesthetic taste in code structure.

ManThursday (approved):
Suggested a couple of API tweaks but lgtm regardless.

packages/browser/src/types/provider.ts
bgiori: Thanks for the review! 

You bring up a good point about how specific this provider should be. With your suggestion this provider would be specific to tracking exposure and additional events we may want to track in the future would need to have their own providers. That being said, there's not a really good example of "additional events we may want to track in the future". The only one I can think of is tracking Assignment events with local-evaluation SDKs. @nirmal91 @curtisliu can you guys think of any other events we would want to track besides exposure and local-evaluation assignment?

My second thought regarding passing an event name and properties is how we are going to make the event "free" (such that it does not count towards the customer's analytics volume). If we are going to filter out events with a certain name from the customer event volume, then passing the event name explicitly in the api may be preferred. 

@curtisliu you had commented on the markdown document about passing the event name and properties rather than user/variant. Would you be able to chime in?

ManThursday: > With your suggestion this provider would be specific to tracking exposure ...

I would just add new methods to the provider rather than a whole new provider in each case.

> That being said, there's not a really good example of "additional events we may want to track in the future".

Yeah it may be incorrect of me to assume that we would want to. SDKs are one of the few areas that it can be worthwhile preempting future requirements but you can still take it too far and make something overly complex that never benefits from that structure.

> If we are going to filter out events with a certain name from the customer event volume, then passing the event name explicitly in the API may be preferred.

Even with a customer provider I would still expect folks to use the Amplitude default provider for the Amplitude part e.g.

```javascript
class CustomTrackingProvider {
  constructor(amplitudeInstance: AmplitudeInstance) {
    this.amplitudeProvider = new AmplitudeTrackingProvider(amplitudeInstance)
  }

  trackExposure(key: string, variant: string): void {
    this.amplitudeProvider.trackExposure(key, variant);

    customTrackingLogic(key, variant);
  }
}
```

nirmal91: Other than exposure and assignment, there aren't strong use cases for other events. We might send some sdk-state-events like "background-flags-updated", "user-state-changed", etc. In the future, we might also want to send heartbeat events to see if there's a socket open here. 

I like `trackExposure` over a generic `track`. If we want to track more things, we can add methods for it later. Basically, experiment should talk to analytics through pre-defined functions. I also prefer us choosing the event-type instead of the customer choosing it. Someone messing it up and having different entries there will be horrible for their Analytics taxonomy. 

Preferred signature:
trackExposure(key, variant) {
   amplitudeInstance.logEvent('[Experiment] Exposure', {'experiment-prefix'+key: variant} // event-props
}

nirmal91 (approved):

packages/browser/src/types/provider.ts
curtisliu: Since we have an exposure event per variant call now, we can use for the properties:
```
{
  "key": ${experimentKey},
  "variant": ${variant}
}
```
in order to reduce number of added event properties

curtisliu: Regarding passing the event name and properties, I think it's pretty important for consistency within Amplitude if customers have something like Segment. I wonder if we should instead pass in an ExposureEvent object that contains:
- key
- variant
- analyticsEventType
- analyticsProperties

ManThursday: The thing that really makes me feel itchy is the idea of having to test the event type string to see what kind of event is being logged e.g.

```javascript
track(eventName, properties) {
  if (eventName === '[Experiment] Exposure') {
    ...
  } else if (...) {
    ...
  }
}
```

Having an explicit `ExposureEvent` class would also solve that in my eyes e.g.

```javascript
track(event) {
  if (event instanceof ExposureEvent) {
    ...
  } else if (...) {
    ...
  }
}
```

I know it's not really any different but it _feels_ more robust.

curtisliu: oh I'm for doing `trackExposure` instead of just `track`

curtisliu: but still prefer passing in an ExposureEvent with an eventName that we control

bgiori: @ManThursday @curtisliu @nirmal91 

I've implemented an event object based api and updated the nomenclature from `tracking` to `analytics`.

Reasons for this approach:
 * Makes it really simple for custom implementation using Amplitude, Segment, or any other analytics SDK by using the base `ExperimentAnalyticsEvent` `name` and `properties` fields.
 * Extensible API for any other events we may want to add in the future. 
     * Allows implementation to opt-in or out of events based on the instance type.
     * `ExperimentAnalyticsEvent` interface implementation can include arbitrary fields.
 * Event name & property consistency across implementations
 
Let me know what you all think.

bgiori: Oh, @curtisliu I missed your comment before implementing. I am worried it is over engineered. 😕

packages/browser/src/config.ts
```diff
@@ -81,6 +100,8 @@ export interface ExperimentConfig {
  | **serverUrl**    | `"https://api.lab.amplitude.com"` |
  | **assignmentTimeoutMillis**    | `10000` |
  | **retryFailedAssignment**    | `true` |
+ | **userProvider**    | `null` |
+ | **trackingProvider**    | `null` |
```
curtisliu, line 104: update comment

curtisliu (approved):

PR #8, opened 2021-07-19
fix: add additional client tests and fixes

bgiori:
### Summary

* more tests for experiment client
* fixes based on tests

### Checklist

### DIFF
--- a/packages/browser/src/experimentClient.ts
+++ b/packages/browser/src/experimentClient.ts
@@ -147,8 +147,12 @@ export class ExperimentClient implements Client {
     if (!this.user) {
       return this.user;
     }
-    const userPropertiesCopy = { ...this.user.user_properties };
-    return { ...this.user, user_properties: userPropertiesCopy };
+    if (this.user?.user_properties) {
+      const userPropertiesCopy = { ...this.user.user_properties };
+      return { ...this.user, user_properties: userPropertiesCopy };
+    } else {
+      return { ...this.user };
+    }
   }
 
   /**
@@ -161,8 +165,12 @@ export class ExperimentClient implements Client {
       this.user = null;
       return;
     }
-    const userPropertiesCopy = { ...user.user_properties };
-    this.user = { ...user, user_properties: userPropertiesCopy };
+    if (this.user?.user_properties) {
+      const userPropertiesCopy = { ...user.user_properties };
+      this.user = { ...user, user_properties: userPropertiesCopy };
+    } else {
+      this.user = { ...user };
+    }
   }
 
   /**
@@ -318,7 +326,7 @@ export class ExperimentClient implements Client {
   private sourceVariants(): Variants {
     if (this.config.source == Source.LocalStorage) {
       return this.storage.getAll();
-    } else if (this.config == Source.InitialVariants) {
+    } else if (this.config.source == Source.InitialVariants) {
       return this.config.initialVariants;
     }
   }
@@ -326,7 +334,7 @@ export class ExperimentClient implements Client {
   private secondaryVariants(): Variants {
     if (this.config.source == Source.LocalStorage) {
       return this.config.initialVariants;
-    } else if (this.config == Source.InitialVariants) {
+    } else if (this.config.source == Source.InitialVariants) {
       return this.storage.getAll();
     }
   }
--- a/packages/browser/test/client.test.ts
+++ b/packages/browser/test/client.test.ts
@@ -66,11 +66,14 @@ test('ExperimentClient.fetch, no retries, timeout failure', async () => {
  */
 test('ExperimentClient.fetch, with retries, retry success', async () => {
   const client = new ExperimentClient(API_KEY, {
+    fallbackVariant: fallbackVariant,
     fetchTimeoutMillis: 1,
   });
   await client.fetch(testUser);
+  let variant = client.variant(serverKey);
+  expect(variant).toEqual(fallbackVariant)
   await delay(2000);
-  const variant = client.variant(serverKey);
+  variant = client.variant(serverKey);
   expect(variant).toEqual(serverVariant);
 });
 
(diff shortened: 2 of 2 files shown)


packages/browser/src/experimentClient.ts
```diff
@@ -147,8 +147,12 @@ export class ExperimentClient implements Client {
     if (!this.user) {
       return this.user;
     }
-    const userPropertiesCopy = { ...this.user.user_properties };
-    return { ...this.user, user_properties: userPropertiesCopy };
+    if (this.user?.user_properties) {
+      const userPropertiesCopy = { ...this.user.user_properties };
+      return { ...this.user, user_properties: userPropertiesCopy };
+    } else {
+      return { ...this.user };
```
ManThursday, line 154: Does `user_properties` change anything?

It's valid to spread a null or undefined variable, you just won't get any properties. So the before and after here are mostly functionally the same.

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,134 @@
...
+
+const delay = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+class TestUserProvider implements ExperimentUserProvider {
+  getUser(): ExperimentUser {
+    return testUser;
+  }
+}
 
 const API_KEY = 'client-DvWljIjiiuqLbyjqdvBaLFfEBrAvGuA3';
```
ManThursday, line 14: These are tests against the live API?

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,134 @@
...
 });
 
-test('ExperimentClient.fetch, no retries, timeout failure', async () => {
-  await testTimeoutRetrySuccessClient.fetch(testUser);
-  const variant = testClient.variant('sdk-ci-test');
+test('ExperimentClient.fetch, with retries, retry success', async () => {
+  const client = new ExperimentClient(API_KEY, {
+    fetchTimeoutMillis: 1,
+  });
+  await client.fetch(testUser);
```
ManThursday, line 56: Does this fetch wait from the retry?

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,134 @@
...
+  await delay(1000);
+  const variant = client.variant('sdk-ci-test');
+  expect(variant).toEqual({ value: 'on', payload: 'payload' });
+});
+
+test('ExperimentClient.fetch, fallbacks returned in correct order', async () => {
```
ManThursday, line 62: In the context of the SDK I have no idea what this is testing.

I would like to see a clear description of the logic being tested step by step in any test with multiple steps.

On top of that this seems like it might be testing the API as much as the SDK?

We should have end-to-end tests somewhere but I'm not sure this suite is the right place for that. Probably a broader discussion to have outside of this PR.

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,134 @@
...
+  expect(variant).toEqual(offVariant);
+});
+
+test('ExperimentClient.fetch, sets user, setUser overrides', async () => {
+  const client = new ExperimentClient(API_KEY, {});
+  await client.fetch(testUser);
+  expect(client.getUser()).toEqual(testUser);
+  const newUser = { user_id: 'new_test_user' };
+  client.setUser(newUser);
+  expect(client.getUser()).toEqual(newUser);
```
ManThursday, line 124: This part seems straight forward. How does doing a fetch before this affect things? Is there a side effect we should also be testing for?

packages/browser/src/experimentClient.ts
bgiori: This was added to make testing easier. Comparing the original user (null/undefined user_properties) and the one as a result of `get/setUser()`would not be equal.

packages/browser/test/client.test.ts
bgiori: Yes. I could mock the response eventually, but its a lot easier to hit the live endpoint.

packages/browser/test/client.test.ts
bgiori: No, the retries occur in the background.

packages/browser/test/client.test.ts
bgiori: `fetch` with a user argument also sets the user within the SDK. This is probably the "feature" I like the least when it comes to using the SDK. 

So this test checks that `fetch` sets the user and that calling `setUser` after the fact overwrites whatever was set in `fetch`

packages/browser/test/client.test.ts
bgiori: This test is mostly testing the order/priority of fallbacks within the SDK--both before and after fetching variants from the server. 

I will add comments on each step describing what we are testing and what we expect.

packages/browser/test/client.test.ts
ManThursday: Could you assert that `variant('sdk-ci-test')` at this point returns the config fallback? I think that would help demonstrate the intention of the test.

bgiori:
@ManThursday could you re-review? Added a comment for each test case & split up the big one.

ManThursday (commented):
Breaking up the big test definitely helped. The descriptions are good as well although I had actually been thinking more along the lines of smaller comments annotating individual steps within the tests

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,194 @@
...
+ * Setting source to initial variants will prioritize variants in initial
+ * variants over those stored in initial variants.
+ */
+test('ExperimentClient.fetch, initial variants source, prefer initial', async () => {
+  const client = new ExperimentClient(API_KEY, {
+    source: Source.InitialVariants,
+    initialVariants: initialVariants,
+  });
+  let variant = client.variant(serverKey);
+  expect(variant).toEqual(serverOffVariant);
```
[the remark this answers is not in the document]
bgiori: That's correct.

packages/browser/test/client.test.ts
```diff
@@ -1,38 +1,194 @@
...
+  await client.fetch(testUser);
```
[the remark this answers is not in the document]
bgiori: Yeah

ManThursday (approved):
lgtm

PR #9, opened 2021-07-25
fix: don't track exposure to fallbacks and add tests

bgiori:
### Summary

* Don't track exposure to fallback variants
* Add tests for analytics provider exposure tracking

### Checklist

* Does your PR have a breaking change?:   NO


ManThursday:
Why shouldn't we track exposure to fallback variants?

bgiori:
> Why shouldn't we track exposure to fallback variants?

I posted about this in #skylab-internal

My guess is that a customer will never use an exposure to a fallback metric as the exposure metric in an experiment. Therefore it is not useful to send. 

For debugging, tracking exposures to fallbacks would be useful. Nirmal brought up that we could send a different metric in this case (e.g. `[Experiment] Fallback Exposure`).

packages/browser/src/experimentClient.ts
```diff
@@ -107,27 +107,39 @@ export class ExperimentClient implements Client {
...
    * @param fallback The highest priority fallback.
    * @see ExperimentConfig
+   * @see ExperimentAnalyticsProvider
    */
   public variant(key: string, fallback?: string | Variant): Variant {
     if (!this.apiKey) {
       return { value: undefined };
     }
+    const sourceVariant = this.sourceVariants()[key];
+    if (sourceVariant?.value) {
```
ManThursday, line 127: Why not `if (sourceVariant) {`?

I had to check but we do currently require a value so this existing code is correct.

bgiori: Technically, sourceVariant could be non-null with a null value, although this is definitely not the regular case and I'd doubt it'd ever happen. That being said, a null value means an invalid variant, which we do not want to track.

ManThursday (approved):
lgtm

PR #11, opened 2021-08-10
feat: add user properties to analytics events like exposure

bgiori:
### Checklist

### DIFF
--- a/packages/browser/src/types/analytics.ts
+++ b/packages/browser/src/types/analytics.ts
@@ -16,11 +16,16 @@ export interface ExperimentAnalyticsEvent {
   name: string;
 
   /**
-   * Properties for the analytics event. Should be passed as the event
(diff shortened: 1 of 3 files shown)


packages/browser/src/types/analytics.ts
```diff
@@ -53,5 +59,6 @@ export class ExposureEvent implements ExperimentAnalyticsEvent {
       key: key,
       variant: variant.value,
     };
+    this.userProperties = user.user_properties;
```
curtisliu, line 62: This should be
```
this.userProperties = {
  `[Experiment] ${key}`: variant.value
}```

curtisliu (approved):

packages/browser/src/types/analytics.ts
curtisliu: may want to pull the string template into a constant

PR #13, opened 2021-10-06
feat: unset user properties when variant evaluates to none or is a fallback

curtisliu:
### Summary

This adds the ability to unset user properties when a variant has no value or is a fallback.

Other changes:
- Fallback order changed for source = InitialVariants:
  1. InitialFlags
  2. Local Storage
  3. Function fallback
  4. Config fallback
- Added `trackExposure` parameter to `variant()` to allow explicitly not tracking exposure. Defaults to true.

### Checklist

* Does your PR have a breaking change?:  Breaking in the sense that custom ExperimentAnalyticsProviders will need to define an `unset` method.

### DIFF
--- a/packages/browser-demo/src/experiment.js
+++ b/packages/browser-demo/src/experiment.js
@@ -23,7 +23,8 @@ const config = {
   },
   analyticsProvider: {
     track: (event) => { console.log('Tracked: ' + JSON.stringify(event))},
-    unset: (event) => { console.log('Unset: ' + JSON.stringify(event))}
+    setUserProperty: (event) => { console.log('Set User Property: ' + JSON.stringify(event))},
+    unsetUserProperty: (event) => { console.log('Unset User Property: ' + JSON.stringify(event))}
   }
 }
 
--- a/packages/browser/src/types/client.ts
+++ b/packages/browser/src/types/client.ts
@@ -8,11 +8,7 @@ import { Variant, Variants } from './variant';
  */
 export interface Client {
   fetch(user?: ExperimentUser): Promise<Client>;
-  variant(
-    key: string,
-    fallback?: string | Variant,
-    trackExposure?: boolean,
-  ): Variant;
+  variant(key: string, fallback?: string | Variant): Variant;
   all(): Variants;
   getUser(): ExperimentUser;
   setUser(user: ExperimentUser): void;


packages/browser/src/types/analytics.ts
```diff
@@ -26,41 +26,70 @@ export interface ExperimentAnalyticsEvent {
...
+  FALLBACK_INLINE = 'fallback-inline',
+  FALLBACK_CONFIG = 'fallback-config',
 }
+
+export const isFallback = (source: VariantSource): boolean => {
+  return (
+    source === VariantSource.FALLBACK_INLINE ||
+    source === VariantSource.FALLBACK_CONFIG
+  );
+};
```
bgiori, line 95: Not sure this belongs in this file.

Could put this and `Source` from config in its own file within `/types`

packages/browser/src/types/provider.ts
```diff
@@ -18,4 +18,5 @@ export interface ExperimentUserProvider {
  */
 export interface ExperimentAnalyticsProvider {
   track(event: ExperimentAnalyticsEvent): void;
+  unset(event: ExperimentAnalyticsEvent): void;
```
bgiori, line 21: This is going to break the smart recruiters custom implementation. We should technically release a v2...

Maybe we discuss the change with them and work through their implementation together? I'd like to avoid publishing v2 until we overhaul the library. 😕

Alternatively, we could have the logic built into the event in track like we do with set user properties. and unset the property if the variant value is null (unassigned or fallback).

packages/browser/src/types/analytics.ts
```diff
@@ -26,41 +26,70 @@ export interface ExperimentAnalyticsEvent {
    * User properties to identify with the user prior to sending the event.
    */
   userProperties?: Record<string, unknown>;
-}
-
-/**
- * Event for tracking a user's exposure to a variant. This event will not count
- * towards your analytics event volume.
- */
-export class ExposureEvent implements ExperimentAnalyticsEvent {
```
bgiori, line 35: Is this pattern not common in js-land? In other words is it more "native" to use a function to return an interface implementation?

packages/browser/test/client.test.ts
```diff
@@ -219,79 +215,64 @@ test('ExperimentClient.fetch, with config user provider, success', async () => {
...
+    }),
+    key: serverKey,
+    variant: serverVariant,
+    userProperties: {
+      [`[Experiment] ${serverKey}`]: serverVariant.value,
+    },
+    userProperty: `[Experiment] ${serverKey}`,
+  });
+
+  expect(spyUnset).toBeCalledTimes(0);
```
bgiori, line 259: Ah this is cool, still such a noob with js testing frameworks.

packages/browser/src/types/analytics.ts
```diff
@@ -26,41 +26,70 @@ export interface ExperimentAnalyticsEvent {
    * User properties to identify with the user prior to sending the event.
    */
   userProperties?: Record<string, unknown>;
```
bgiori, line 28: Since you added `userProperty` you dont need this anymore?

I am confused that we need both userProperty and userProperties.

packages/browser/src/types/analytics.ts
curtisliu: for simple data objections, factory functions are preferable to classes

packages/browser/src/types/analytics.ts
curtisliu: Yeah, this isn't necessary anymore but keeping it for backwards compatibility

packages/browser/src/types/analytics.ts
curtisliu: 👍

packages/browser/src/types/provider.ts
curtisliu: I made this function optional, which should make it backwards compatible

packages/browser/src/experimentClient.ts
```diff
@@ -114,37 +115,122 @@ export class ExperimentClient implements Client {
...
    * @param fallback The highest priority fallback.
+   * @param trackExposure Sends a track event via the configured {@link ExperimentAnalyticsProvider}
    * @see ExperimentConfig
    * @see ExperimentAnalyticsProvider
    */
-  public variant(key: string, fallback?: string | Variant): Variant {
+  public variant(
+    key: string,
+    fallback?: string | Variant,
+    trackExposure = true,
```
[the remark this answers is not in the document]
curtisliu: Main reason for this is to allow them to call variant multiple times, then call it once when they actually want to send the event. Maybe an explicit track exposure call is the right interface, so I'll remove it from this PR

curtisliu:
@bgiori PTAL

PR #15, opened 2021-11-18
feat: core package for seamless integration with amplitude analytics

bgiori:
### Summary

Add core package for integration with analytics SDK.

 * Keeps the same API as Android/iOS for `AnalyticsConnector` and `IdentityStore`
      * Changes the core singleton access to global/window in order to play nicely in the browser with potentially different script tags.
 * Package is pulled in as a dependency to experiment SDK and a dev dependency to the analytics SDK.

### Checklist

* Does your PR have a breaking change?:   no


bgiori:
This is not quite done yet, refactoring the component architecture.

packages/core/src/amplitudeCore.ts
```diff
@@ -0,0 +1,17 @@
...
+import { safeGlobal } from './util/global';
+
+safeGlobal['amplitudeCoreInstances'] = {};
+
+export class AmplitudeCore {
+  public readonly identityStore = new IdentityStoreImpl();
+  public readonly analyticsConnector = new AnalyticsConnectorImpl();
+
+  static getInstance(instanceName: string): AmplitudeCore {
+    if (!safeGlobal['amplitudeCoreInstances'][instanceName]) {
```
qingzhuozhen, line 12: Is this to resolve earlier issue mentioned ?

packages/core/rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+
+import tsConfig from './tsconfig.json';
+
+const browserConfig = {
+  input: 'src/index.ts',
+  output: {
+    dir: 'dist',
+    entryFileNames: 'core.umd.js',
+    exports: 'named',
+    format: 'umd',
```
qingzhuozhen, line 16: Curious is esm also supported?

packages/core/src/analyticsConnector.ts
```diff
@@ -0,0 +1,35 @@
...
+  setEventReceiver(listener: AnalyticsEventReceiver): void;
+}
+
+export class AnalyticsConnectorImpl implements AnalyticsConnector {
+  private receiver: AnalyticsEventReceiver;
+  private queue: AnalyticsEvent[] = [];
+
+  logEvent(event: AnalyticsEvent): void {
+    if (!this.receiver) {
+      this.queue.push(event);
```
justin-fiedler, line 20: The queue could grow unbounded if no receiver is added. Maybe we could add a max queue size or user warning is no receiver is set.

justin-fiedler, line 20: The queue could grow unbounded if no receiver is added. Maybe we could add a max queue size or user warning is no receiver is set.

packages/core/src/amplitudeCore.ts
bgiori: Yes, even if the same code is copied and used separately, the core instance is always a truly global singleton.

packages/core/rollup.config.js
bgiori: I need to add support for this later. High priority ATM but not blocking this release.

Issue #37, opened 2022-07-11
please publish an esm version of experiments-js-client & analytics-connector


sod:
please publish  `@amplitude/experiments-js-client` & `@amplitude/analytics-connector` both in an esm version (just like you already do with `amplitude-js` & `@amplitude/ua-parser-js`), so webpack can optimize it. 

Right now our amplitude chunk looks like this:

![CleanShot 2022-07-11 at 16 38 37](https://user-images.githubusercontent.com/905328/178290117-5a4af2fb-c72f-4333-8b3f-e3f6802df226.png)

It's ~122kb of amplitude stuff, when it could be ~89kb with code sharing.

Right now, dependencies are:

* `amplitude-js`
  * imports `@amplitude/ua-parser-js`
  * imports `@amplitude/analytics-connector`
* `@amplitude/analytics-connector` - which is right now umd only, and imports are inlined:
  * imports `@amplitude/ua-parser-js` (inlined)
* `@amplitude/experiments-js-client` - which is right now umd only, and imports are inlined:
  * imports `@amplitude/analytics-connector` (inlined)
    * imports `@amplitude/ua-parser-js` (inlined)

so as we have to use `@amplitude/experiments-js-client` + `amplitude-js`, webpack is forced to include:
* 3 times `@amplitude/ua-parser-js`
* 2 times `@amplitude/analytics-connector`
* 1 times `@amplitude/experiments-js-client`
* 1 times `amplitude-js`

PR #38, opened 2022-07-11
build: add esm and es2015 builds

sod:
fixes https://github.com/amplitude/experiment-js-client/issues/37

* keep existing umd build untouched to not break existing usages
* add `{ "module": "dist/<package>.esm.js" }` with `es5` syntax and es modules
* add `{ "es2015": "dist/<package>.es2015.js" }` with `es2015` syntax and es modules

So you can reuse modules from amplitude (instead of e.g. having `@amplitude/ua-parser-js` 3 times in your build) and if you don't care about IE11, can use the smaller es2015 syntax.

CleanShot 2022-07-11 at 19 02 56@2x

### DIFF
--- a/babel.es2015.config.js
+++ b/babel.es2015.config.js
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          browsers: ['chrome 10'],
+        },
+      },
+    ],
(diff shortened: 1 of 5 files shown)


bgiori:
Thanks @sod!

bgiori (approved):
Looks good to me!

Just going to get one other review before merge.

kevinpagtakhan (approved):
💯 

Thanks @sod!

PR #47, opened 2022-10-12
docs: experiment examples - html

bgiori:
### Summary

New html script tag examples for experiment SDKs

### Checklist

### DIFF
--- a/examples/html-app/amplitude-integration/index.html
+++ b/examples/html-app/amplitude-integration/index.html
@@ -11,38 +11,18 @@
     Use the SDK hosted on unpkg.com to avoid ad blockers blocking downloading the SDK.
   -->
   <script type="text/javascript">
-    (function(e,t){var r=e.amplitude||{_q:[],_iq:{}};var n=t.createElement("script")
-    ;n.type="text/javascript"
-    ;n.crossOrigin="anonymous";n.async=true
-    ;n.src="https://unpkg.com/amplitude-js@8.18.4/amplitude.umd.min.js"
-    ;n.onload=function(){if(!e.amplitude.runQueuedFunctions){
-      console.log("[Amplitude] Error: could not load SDK")}}
-    ;var s=t.getElementsByTagName("script")[0];s.parentNode.insertBefore(n,s)
-    ;function i(e,t){e.prototype[t]=function(){
-      this._q.push([t].concat(Array.prototype.slice.call(arguments,0)));return this}}
-      var o=function(){this._q=[];return this}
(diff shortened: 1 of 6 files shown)


qingzhuozhen (approved):
looks great

examples/html-app/amplitude-integration/index.html
```diff
@@ -0,0 +1,66 @@
...
... s=t.getElementsByTagName("script")[0];function v(e,t){e.prototype[t]=function(){return this._q.push({name:t,args:Array.prototype.slice.call(arguments,0)}),this}}s.parentNode.insertBefore(n,s);for(var o=function(){return this._q=[],this},i=["add","append","clearAll","prepend","set","setOnce","unset","preInsert","postInsert","remove","getUserProperties"],a=0;a<i.length;a++)v(o,i[a]);r.Identify=o;for(var u=function(){return this._q=[],this},c=["getEventProperties","setProductId","setQuantity","setPrice","setRevenue","setRevenueType","setEventProperties"],l=0;l<c.length;l++)v(u,c[l]);r.Revenue=u;var p=["getDeviceId","setDeviceId","getSessionId","setSessionId","getUserId","setUserId","setOptOut","setTransport","reset"],d=["init","add","remove","track","logEvent","identify","groupIdentify","setGroup","revenue","flush"];function f(e){function t(t,r){e[t]=function(){var n={promise:new Promise((r=>{e._q.push({name:t,args:Array.prototype.slice.call(arguments,0),resolve:r})}))};if(r)return n}}for(var r=0;r<p.length;r++)t(p[r],!1);for(var n=0;n<d.length;n++)t(d[n],!0)}f(r),r.createInstance=function(){var e=r._iq.push({_q:[]})-1;return f(r._iq[e]),r._iq[e]},e.amplitude=r}}(window,document)}();
```
kevinpagtakhan, line 15: Curious why we used the script from npm instead of the one from s3?

kevinpagtakhan:
nit: should the commit be `docs` instead?

examples/html-app/amplitude-integration/index.html
bgiori: Adblockers block `cdn.amplitude.com` which makes it so that we cant event initialze the SDK. This makes it impossible for the expeirment SDK to set up the integration from `initializeWithAmplitudeAnalytics()` and the fetch() request will wait an timeout after 10 seconds. If the customer is awaiting the result of `fetch()` to load their UI, its better for SDK to get initialized and the `fetch()` request from the SDK to be blocked immediately, rather than delay loading the UI until the 10sec timeout is finished. 

Unpkg also uses a CDN to cache these packages, so I think the load time should not be too different.

PR #68, opened 2023-04-14
fix(deps): Use a range constraint for the ua-parser package

stof:
### Summary

Given that the ua-parser package does not live in this mono-repo, its releases are not in sync. The exact version used here leads to installing 2 different versions of the ua-parser as part of the dependencies of the analytics-browser package, and both are part of the bundle, increasing the bundle size.

Closes https://github.com/amplitude/Amplitude-TypeScript/issues/373

### Checklist


bgiori:
Hi @stof, thanks for this contribution! Sorry for the delay.

PR #70, opened 2023-05-08
feat: remove @analytics/ua-parser as dependency of @amplitude/analyti…

kevinpagtakhan:
### Summary

* Removes `@analytics/ua-parser` as dependency of @amplitude/analytics-connector
* Adds `@analytics/ua-parser` as dependency of @amplitude/experiment-js-client

The ultimate goal is to remove `@analytics/ua-parser` as direct/indirect dependency of `@amplitude/analytics-browser`.

### Checklist

### DIFF
--- a/packages/experiment-browser/rollup.config.js
+++ b/packages/experiment-browser/rollup.config.js
@@ -71,7 +71,7 @@ const configs = [
       entryFileNames: 'experiment.esm.js',
       format: 'esm',
     }),
-    external: ['@amplitude/ua-parser-js', '@amplitude/analytics-connector'],
+    external: ['@amplitude/analytics-connector'],
   },
 
   // modern build for field "es2015" - not ie, esm, es2015 syntax
@@ -81,7 +81,7 @@ const configs = [
       entryFileNames: 'experiment.es2015.js',
       format: 'esm',
     }),
-    external: ['@amplitude/ua-parser-js', '@amplitude/analytics-connector'],
+    external: ['@amplitude/analytics-connector'],
   },
 ];
 
--- a/packages/analytics-connector/package.json
+++ b/packages/analytics-connector/package.json
@@ -27,9 +27,7 @@
   "bugs": {
     "url": "https://github.com/amplitude/experiment-js-client/issues"
   },
-  "dependencies": {
-    "@amplitude/ua-parser-js": "^0.7.31"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@types/amplitude-js": "^8.0.2",
     "amplitude-js": "^8.12.0"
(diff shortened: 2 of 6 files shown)


nirmal91:
for just the analytics sdk, if we don't use ua-parser, how are we going to get os and device model?

liuyang1520:
> for just the analytics sdk, if we don't use ua-parser, how are we going to get os and device model?

Same question, are we going to rely on the pipeline to determine the device info for analytics sdk?

kevinpagtakhan:
@nirmal91 
> we don't use ua-parser, how are we going to get os

@liuyang1520 
> Same question, are we going to rely on the pipeline to determine the device info for analytics sdk?

The Browser SDK will send raw user agent values from `navigator.userAgent` and let ingestion parse and populate the respective fields which I believe happens here: https://github.com/amplitude/amplitude/blob/master/event/converters/http_event.py#L353-L367

liuyang1520:
> @nirmal91
> 
> > we don't use ua-parser, how are we going to get os
> 
> @liuyang1520
> 
> > Same question, are we going to rely on the pipeline to determine the device info for analytics sdk?
> 
> The Browser SDK will send raw user agent values from `navigator.userAgent` and let ingestion parse and populate the respective fields which I believe happens here: https://github.com/amplitude/amplitude/blob/master/event/converters/http_event.py#L353-L367

Thanks @kevinpagtakhan for sharing the details! Should we also rely on the pipeline/ingestion side for experiment to get consistent parsed results? Probably out of scope for this PR though, changes LGTM.

liuyang1520 (approved):

qingzhuozhen (approved):

packages/experiment-browser/rollup.config.js
```diff
@@ -71,7 +71,7 @@ const configs = [
       entryFileNames: 'experiment.esm.js',
       format: 'esm',
     }),
-    external: ['@amplitude/ua-parser-js', '@amplitude/analytics-connector'],
+    external: ['@amplitude/analytics-connector'],
```
stof, line 74: shouldn't it stay an external here given that experiment-browser still depend on it ? bundling the ua-parser-js code would prevent deduplicating it with other packages using it, which hurts the "tree shakable" goal of that build.

kevinpagtakhan: hey @stof, the thought here is that other packages that used to depend on it like , no longer do like @amplitude/analytics-connector, @amplitude/analytics-browser. our ultimate goal is to parse user agent server side so the SDKs no longer depend on the client side user agent parser. lmk if you have further thoughts!

stof: Anything that is in the `dependencies` section of the npm package should actually be listed in externals, so that they get imported from the installed dependency (allowing bundlers to dduplicate things).

kevinpagtakhan: @stof gotcha! i see what you mean now. i realized that our esm build methods for analytics and experiment SDKs are different. i will open a pull request.

PR #88, opened 2023-09-18
fix: optimize variant cache access, legacy exposure

bgiori:
### Summary

* Call `get()` for accessing the variant rather than `getAll()[key]`
* Check if there is a legacy exposure tracker before adding context to the user

### Checklist


paladin8 (approved):
Thanks for the quick fix! I assume we'll release a new version and then upgrade the client used in our app?

bgiori:
> Thanks for the quick fix! I assume we'll release a new version and then upgrade the client used in our app?

Yep, thanks for the review, and thanks again for finding this.

PR #97, opened 2024-02-01
feat: Enable use of project API key for default deployments

tyiuhc:
### Summary

Enable use of project API key for default deployments

### Checklist

### DIFF
--- a/packages/experiment-browser/src/config.ts
+++ b/packages/experiment-browser/src/config.ts
@@ -133,6 +133,11 @@ export interface ExperimentConfig {
    * (Advanced) Use your own http client.
    */
   httpClient?: HttpClient;
+
+  /**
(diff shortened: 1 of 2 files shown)


packages/experiment-browser/src/config.ts
```diff
@@ -180,4 +186,5 @@ export const Defaults: ExperimentConfig = {
   analyticsProvider: null,
   exposureTrackingProvider: null,
   httpClient: FetchHttpClient,
+  deploymentKey: null,
```
bgiori, line 189: use undefined from here on out.

bgiori: `null` is a bad practice in javascript IMO. previous null values can stick around but moving forward always prefer undefined.

PR #98, opened 2024-02-27
feat: add stream api

zhukaihan:
### Summary

Adds a stream API so that we can now stream server-side events to SDK's. 
Adds a stream flag API that uses stream SPI so that we can now stream flags to SDK's. 

The flow of execution is complex due to the nature of async everything related to SSE. 
Some [doc](https://amplitude.atlassian.net/l/cp/QnjLxRa6) that may help out understanding the flows. 

### Checklist


packages/experiment-core/src/api/stream-api.ts
```diff
@@ -0,0 +1,194 @@
...
+      Authorization: `Api-Key ${this.deploymentKey}`,
+    };
+    if (options?.libraryName && options?.libraryVersion) {
+      headers[
+        'X-Amp-Exp-Library'
+      ] = `${options.libraryName}/${options.libraryVersion}`;
+    }
+
+    // Create connection. It starts connection on new.
+    const es = new this.eventSourceClass(this.serverUrl, {
```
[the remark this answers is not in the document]
bgiori: I dont think `EventSourceClass` really describes what this is.

Maybe something like `StreamEventSourceConnectionFactory` and `StreamEventSourceConnection` rather than `StreamEventSource` and `StreamEventSourceClass`

That is just a thought. My suggestions are a little wordy but my point is that really the "class" is creating the connection which is used until it is closed. "Factory" might not be the right suffix here, and I'm still not sold on shoehorning in `new` instead of using an explicit interface with a function that creates the connection using the underlying implementation.

packages/experiment-core/src/api/stream-api.ts
```diff
@@ -0,0 +1,194 @@
...
+      } catch {} // Don't care about errors after handoff.
+    }
+  }
+
+  private async handleNewMsg(response: StreamMessageEvent) {
+    if (!response.data) {
+      return;
+    }
+    this.setKeepAliveExpiry(); // Reset keep alive as there is data.
+    if (KEEP_ALIVE_INTERVAL > 0 && response.data == KEEP_ALIVE_DATA) {
```
bgiori, line 168: `KEEP_ALIVE_INTERVAL` is never <= 0

was the idea to make this configurable?

packages/experiment-core/src/api/stream-api.ts
```diff
@@ -0,0 +1,194 @@
...
+      clearTimeout(this.keepAliveTimeout);
+      this.keepAliveTimeout = undefined;
+    }
+    if (KEEP_ALIVE_INTERVAL <= 0) {
+      return;
+    }
+    if (this.eventSource) {
+      this.keepAliveTimeout = setTimeout(() => {
+        this.error(DefaultStreamErrorEvents.KEEP_ALIVE_FAILURE);
+      }, KEEP_ALIVE_INTERVAL);
```
bgiori, line 191: Does the server define the keepalive interval? Is that set to 15 seconds? So this is 15 + 2 for buffer?

packages/experiment-core/src/api/stream-api.ts
```diff
@@ -0,0 +1,194 @@
...
+      } catch {} // Don't care about errors after handoff.
+    }
+  }
+
+  private setKeepAliveExpiry() {
+    if (this.keepAliveTimeout) {
+      clearTimeout(this.keepAliveTimeout);
+      this.keepAliveTimeout = undefined;
+    }
+    if (KEEP_ALIVE_INTERVAL <= 0) {
```
bgiori, line 185: never true

bgiori (commented):
Only reviewed the StreamApi

Reviews for other files, etc still coming.

packages/experiment-core/src/api/stream-api.ts
```diff
@@ -0,0 +1,194 @@
...
+const KEEP_ALIVE_DATA = ' ';
+
+export type StreamOptions = {
+  libraryName: string;
+  libraryVersion: string;
+};
+export type StreamOnUpdateCallback = (data: string) => unknown;
+export type StreamOnErrorCallback = (err: StreamErrorEvent) => unknown;
+
+export interface StreamApi {
```
bgiori, line 22: Also, one-off: This seems like a transform interface and implementation to me, not necessarily an API.

I would consider moving this into the `transport/streatm.ts` file.

packages/experiment-core/src/api/stream-api.ts
zhukaihan: Yes. server defines keepalive interval. I was thinking about making it configurable, but since it's server-defined, I guess there is no need for configuring it. I'll delete the checks.

packages/experiment-core/src/api/stream-api.ts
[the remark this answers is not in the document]
zhukaihan: I was torned between factory pattern or just passing a class. Chose passing a class to keep the interface similar to other APIs, but factory pattern does makes more sense here.

packages/experiment-core/src/transport/stream.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * This is the type for the class of StreamEventSource.
+ * It should support new operation along with defined static members.
+ */
+export type StreamEventSourceClass = Class<StreamEventSource> & {
+  // These are static members usable without new.
+  readonly CLOSED: number;
+  readonly CONNECTING: number;
+  readonly OPEN: number;
+};
```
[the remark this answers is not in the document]
zhukaihan: Interface with 3 fields requires instantiating a new object to access them, similar to instance variables. having this on type is similar to static variables. 
Was trying to limit access to those on the type instead of objects. Since we may use a factory function to create them, they can just access them using the objects.

zhukaihan:
Moving files to `experiment-node-server` SDK. Currently, only server should support stream and this change reduces client SDK package size.

PR #116, opened 2024-08-12
feat: add web user targeting

zhukaihan:
### Summary

Add user targeting for web experimentation. 

Sets `first_seen` on user object if doesn't exist and save to local storage. 
Sets a number of fields on user object for user targeting. 

### Checklist


bgiori (changes requested):
Lets add this code to the main SDK actually, so we get this behavior OOTB for feature experiment too.

packages/experiment-tag/src/experiment.ts
```diff
@@ -18,16 +19,27 @@ import {
   concatenateQueryParamsOf,
 } from './util';
 
+type WebExpUser = ExperimentUser & {
+  first_seen?: string;
+  device_type?: 'mobile' | 'tablet' | 'desktop';
```
bgiori, line 24: `device_type` is overloaded with the amplitude property. I would recommend changing this property name to be independent. maybe `device_category`?

packages/experiment-tag/src/experiment.ts
```diff
@@ -16,18 +17,25 @@ import {
...
 } from './util';
 
 const appliedMutations: MutationController[] = [];
 let previousUrl: string | undefined = undefined;
 
+type WebExpUser = ExperimentUser & {
+  landing_url?: string;
+  first_seen?: string;
+};
+
```
bgiori, line 30: Sorry, maybe it wasn't clear, we can put this into the main sdk also?

bgiori (changes requested):
More into the main SDK!

bgiori (approved):

PR #117, opened 2024-08-19
feat: Page targeting for Web Experimentation

tyiuhc:
### Summary

Add page targeting functionality for web experimentation

### Checklist

### DIFF
--- a/packages/analytics-connector/src/eventBridge.ts
+++ b/packages/analytics-connector/src/eventBridge.ts
@@ -13,55 +13,43 @@ export type AnalyticsEventReceiver = (event: AnalyticsEvent) => void;
 
 export interface EventBridge {
   logEvent(event: AnalyticsEvent): void;
+
   setEventReceiver(listener: AnalyticsEventReceiver): void;
-  setApiKey(apiKey: string): void;
+
+  setInstanceName(instanceName: string): void;
 }
 
 export class EventBridgeImpl implements EventBridge {
-  private apiKey = '';
+  private instanceName = '';
   private receiver: AnalyticsEventReceiver;
   private inMemoryQueue: AnalyticsEvent[] = [];
   private globalScope = getGlobalScope();
 
   private getStorageKey(): string {
-    return `EXP_unsent_${this.apiKey.slice(0, 10)}`;
-  }
-
-  private initializeStorage(): void {
-    if (isLocalStorageAvailable()) {
-      const storageKey = this.getStorageKey();
-      const storedQueue = this.globalScope.localStorage.getItem(storageKey);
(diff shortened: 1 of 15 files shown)


packages/experiment-tag/src/experiment.ts
```diff
@@ -61,13 +64,25 @@ export const initializeExperiment = (apiKey: string, initialFlags: string) => {
...
+            isPageTargetingSegment(segment),
+          );
+
+          // Create or update the preview segment
+          const previewSegment = {
+            metadata: { segmentName: 'preview' },
+            variant: urlParams[flag.key],
+          };
+
+          flag.segments = [...pageTargetingSegments, previewSegment];
```
bgiori, line 78: I have a feeling our initial implementation is too complicated. Maybe we just force apply the specific flag & variant actions.

bgiori: For another time though. Just a thought

packages/analytics-connector/src/eventBridge.ts
```diff
@@ -9,16 +14,58 @@ export type AnalyticsEventReceiver = (event: AnalyticsEvent) => void;
...
+    } else {
+      this.inMemoryQueue = queue;
+    }
+  }
 
   logEvent(event: AnalyticsEvent): void {
     if (!this.receiver) {
-      if (this.queue.length < 512) {
-        this.queue.push(event);
+      this.initializeStorage(); // Ensure storage is initialized
```
bgiori, line 64: Calling this we are always calling getItem on local storage. Once we initialize the storage, we should store that info in memory in a boolean and only initialize the storage if we have not already in this session.

packages/analytics-connector/src/eventBridge.ts
```diff
@@ -9,16 +14,58 @@ export type AnalyticsEventReceiver = (event: AnalyticsEvent) => void;
...
 
 export class EventBridgeImpl implements EventBridge {
+  private apiKey = '';
   private receiver: AnalyticsEventReceiver;
-  private queue: AnalyticsEvent[] = [];
+  private inMemoryQueue: AnalyticsEvent[] = [];
+  private globalScope = getGlobalScope();
+
+  private getStorageKey(): string {
+    return `EXP_unsent_${this.apiKey.slice(0, 10)}`;
```
bgiori, line 27: I dont think we should use the apikey here. Could just use the instance name which is associated with the analytics Connector that initializes this. 

Right now there's some issues with apiKey in this class. What if we intiailize two instances with the same instance name and different api keys?

bgiori (changes requested):

bgiori (approved):

PR #118, opened 2024-08-20
fix: add inject support to experiment-tag

bgiori:
### Checklist

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -212,10 +212,10 @@ const handleInject = (action, key: string, variant: Variant) => {
     const rawCss = action.data.css;
     let style: HTMLStyleElement | undefined;
     if (rawCss) {
-      style = document.createElement('style');
+      style = globalScope.document.createElement('style');
       style.innerHTML = rawCss;
       style.id = `css-${id}`;
-      document.head.appendChild(style);
+      globalScope.document.head.appendChild(style);
     }
     // Create HTML
     const rawHtml = action.data.html;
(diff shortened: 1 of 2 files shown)


packages/experiment-tag/src/experiment.ts
```diff
@@ -186,6 +189,59 @@ const revertMutations = () => {
...
+};
+
+const handleInject = (action, key: string, variant: Variant) => {
+  const globalScope = getGlobalScope();
+  if (!globalScope) {
+    return;
+  }
+  const urlExactMatch = variant?.metadata?.['urlMatch'] as string[];
+  const currentUrl = urlWithoutParamsAndAnchor(globalScope.location.href);
+  if (matchesUrl(urlExactMatch, currentUrl)) {
```
bgiori, line 205: @tyiuhc I think this is going to conflict with your PR?

tyiuhc: Yes, with page targeting, we would be able to forgo the matchUrl condition. We could merge your PR first, then I can update the logic and tests with page targeting.

packages/experiment-tag/src/experiment.ts
```diff
@@ -186,6 +189,59 @@ const revertMutations = () => {
   }
 };
 
+const inject = (js: string, html, utils, id) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  eval(js, html, utils, id);
```
bgiori, line 195: pass these fields to eval to make it so they're not stripped out by the transpiler

zhukaihan (approved):

PR #121, opened 2024-08-22
chore: use inline inject instead of eval

bgiori:
### Summary

Replace eval with an inline inject

### Checklist

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -246,6 +246,7 @@ const handleInject = (action, key: string, variant: Variant) => {
       fn(html, utils, id);
     }
   } catch (e) {
+    script?.remove();
     console.error(
       `Experiment inject failed for ${key} variant ${variant.key}. Reason:`,
       e,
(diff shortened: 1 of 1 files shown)


packages/experiment-tag/src/experiment.ts
```diff
@@ -202,11 +196,26 @@ const handleInject = (action, key: string, variant: Variant) => {
...
-    const id = action.data.id;
     if (appliedInjections.has(id)) {
       return;
     }
+    // Create JS
+    const rawJs = action.data.js;
+    let script: HTMLScriptElement | undefined;
+    if (rawJs) {
+      script = globalScope.document.createElement('script');
+      script.innerHTML = `function ${id}(html, utils, id){${rawJs}};`;
```
zhukaihan, line 215: Perhaps `script?.xxx` as script can take undefined type? Or `if (rawJs && script)`.

packages/experiment-tag/src/experiment.ts
```diff
@@ -226,14 +235,24 @@ const handleInject = (action, key: string, variant: Variant) => {
...
-    const js = action.data.js;
     appliedInjections.add(id);
-    inject(js, html, utils, id);
+    try {
+      const fn = globalScope[id];
+      if (fn && typeof fn === 'function') {
+        fn(html, utils, id);
+      }
+    } catch (e) {
+      console.error(
```
zhukaihan, line 245: Do we want to remove the script and style html tag elements from document if inject failed?

zhukaihan (approved):

tyiuhc (approved):

packages/experiment-tag/src/experiment.ts
bgiori: Yeah we should probably remove the script. Style may still be useful though.

PR #124, opened 2024-08-27
feat: add url param targeting

zhukaihan:
### Summary

Added URL search parameter targeting. 
Fixed multiple search parameter with same key doesn't get passed to redirect. 
Added global scope to lots of places.

### Checklist

### DIFF
--- a/packages/experiment-tag/src/util.ts
+++ b/packages/experiment-tag/src/util.ts
@@ -79,9 +79,7 @@ export const concatenateQueryParamsOf = (
   const urlObj = new URL(redirectUrl);
 
   globalUrlObj.searchParams.forEach((value, key) => {
-    if (!urlObj.searchParams.has(key)) {
-      urlObj.searchParams.set(key, value);
-    }
+    urlObj.searchParams.append(key, value);
   });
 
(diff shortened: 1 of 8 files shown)


packages/experiment-tag/src/util.ts
```diff
@@ -79,9 +79,7 @@ export const concatenateQueryParamsOf = (
   const urlObj = new URL(redirectUrl);
 
   globalUrlObj.searchParams.forEach((value, key) => {
-    if (!urlObj.searchParams.has(key)) {
-      urlObj.searchParams.set(key, value);
-    }
+    urlObj.searchParams.append(key, value);
```
zhukaihan, line 82: Would like to double check if this change would break anything related to redirects? @tyiuhc

tyiuhc: this should be fine 👍🏻

zhukaihan: Oh yes, going to fix this. 
Changing the behavior to be everything from redirect url plus everything in original url that doesn't conflict with redirect url.

tyiuhc (approved):

PR #132, opened 2024-10-16
fix: Dedupe web experiment exposures and variant actions if URL is unchanged

tyiuhc:
### Summary

 Dedupe web experiment exposures and variant actions if URL is unchanged

### Checklist


packages/experiment-tag/src/experiment.ts
```diff
@@ -22,13 +22,18 @@ import {
...
 
 const appliedInjections: Set<string> = new Set();
 const appliedMutations: MutationController[] = [];
 let previousUrl: string | undefined = undefined;
+// Cache to track exposure for the current URL, should be cleared on URL change
+let urlExposureCache: { [url: string]: { [key: string]: string | undefined } };
 
 export const initializeExperiment = (apiKey: string, initialFlags: string) => {
+  // Clear the cache on initialization
+  urlExposureCache = {};
```
bgiori, line 36: Why? If anything this should be below the check for globalScope.webExperiment

packages/experiment-tag/src/experiment.ts
```diff
@@ -314,13 +318,14 @@ export const setUrlChangeListener = () => {
 
     // Wrapper for replaceState
     history.replaceState = function (...args) {
-      previousUrl = globalScope.location.href;
```
bgiori, line 317: Can you explain this change to me? By moving this to below the apply, we've changed the behavior of the redirect logic in applyRedirect. Since previousUrl will not have been updated.

packages/experiment-tag/src/experiment.ts
```diff
@@ -314,13 +318,14 @@ export const setUrlChangeListener = () => {
...
       // Call the original replaceState
       const result = originalReplaceState.apply(this, args);
-      // Revert mutations and apply variants after replacing state
-      revertMutations();
-      applyVariants(globalScope.webExperiment.all());
-
+      // Revert mutations and apply variants if the URL has changed
+      if (previousUrl && previousUrl !== globalScope.location.href) {
```
bgiori, line 324: Do we also want to make this check for pushState?

packages/experiment-tag/src/experiment.ts
tyiuhc: This was to take care of some unexpected behavior in unit tests.

packages/experiment-tag/src/experiment.ts
tyiuhc: The previous logic was inaccurate - it should've been updated afterwards.

bgiori (approved):

tyiuhc:
Update: this PR has been [altered](https://github.com/amplitude/experiment-js-client/commit/6620cbe34e5779820a309b01285e60d64cf297ce) to not dedupe variant actions to avoid unintended behaviors

PR #138, opened 2024-11-18
feat: Web experiment remote evaluation

tyiuhc:
### Summary

Add remote evaluation support for web experimentation script tag.

### Checklist


packages/experiment-core/src/api/flag-api.ts
```diff
@@ -7,8 +9,13 @@ export type GetFlagsOptions = {
   evaluationMode?: string;
   timeoutMillis?: number;
 };
+
 export interface FlagApi {
-  getFlags(options?: GetFlagsOptions): Promise<Record<string, EvaluationFlag>>;
+  getFlags(
+    options?: GetFlagsOptions,
+    user?: Record<string, unknown>,
+    deliveryMethod?: string | undefined,
```
bgiori, line 17: These should be options.

packages/experiment-tag/src/experiment.ts
```diff
@@ -51,14 +55,21 @@ export const initializeExperiment = (apiKey: string, initialFlags: string) => {
...
-    );
+  // create new user if it does not exist, or it does not have device_id or web_exp_id
+  if (Object.keys(user).length === 0 || !user.device_id || !user.web_exp_id) {
+    if (!user.device_id || !user.web_exp_id) {
+      // if user has device_id, migrate it to web_exp_id
+      if (user.device_id) {
+        user.web_exp_id = user.device_id;
+      } else {
+        const uuid = UUID();
+        user = { device_id: uuid, web_exp_id: uuid };
```
bgiori, line 66: since we call `setUser` with this user, we will never use the device ID passed from the experiment integration (analytics SDK). I think we only want to store the `web_exp_id` in the user, right? Or am I missing something?

packages/experiment-tag/src/experiment.ts
```diff
@@ -121,14 +169,34 @@ export const initializeExperiment = (apiKey: string, initialFlags: string) => {
...
+  if (remoteFlagKeys.size === 0) {
+    return;
+  }
+
+  try {
+    await globalScope.webExperiment.doFlags();
+    // apply remote variants
+    applyVariants(globalScope.webExperiment.all(), remoteFlagKeys);
+  } catch (error) {
+    console.warn('Error fetching remote flags:', error);
```
bgiori, line 191: I think we should still apply local variants in the case that the fetch fails. What's your thought here?

packages/experiment-browser/src/experimentClient.ts
```diff
@@ -701,13 +701,24 @@ export class ExperimentClient implements Client {
     return variants;
   }
 
-  private async doFlags(): Promise<void> {
+  public async doFlags(): Promise<void> {
```
bgiori, line 704: can we keep this private and just `// ts-ignore` the call in experiment-tag?

packages/experiment-tag/src/experiment.ts
tyiuhc: Meaning to evaluate the remote flags with local properties? I would avoid just to prevent variant jumping.

PR #152, opened 2025-01-02
feat: WebExperiment class

tyiuhc:
### Checklist


packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
+  /**
+   * Set the context for evaluating experiments.
+   * If user is not provided, the current user is used.
+   * If currentUrl is not provided, the current URL is used.
+   * @param user
+   * @param currentUrl
+   */
+  public setContext(
+    user: ExperimentUser | undefined = undefined,
+    currentUrl: string | undefined = undefined,
```
bgiori, line 250: We should make this an object `WebExperimentContext` or something where each of these are optional fields.

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
-
-const applyVariants = (
-  variants: Variants,
-  flagKeys: Set<string> | undefined = undefined,
-) => {
-  if (Object.keys(variants).length === 0) {
-    return;
+
+  /**
+   * Set the previous URL for tracking back/forward navigation. Set previous URL to prevent infinite redirection loop
```
bgiori, line 273: Is there a way we can prevent this for them?

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
-  // Initialize the cache if on a new URL
-  if (!urlExposureCache?.[currentUrl]) {
-    urlExposureCache = {};
-    urlExposureCache[currentUrl] = {};
+
+  /**
+   * Revert mutations applied by the experiment.
+   * @param flagKeys
+   */
+  public revertMutations(flagKeys: string[] | undefined = undefined) {
```
bgiori, line 329: Change this to `revertVariants` to match `applyVariants` in verbage.

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
   }
-  const globalScope = getGlobalScope();
-  if (!globalScope) {
-    return;
+
+  /**
+   * Apply evaluated variants to the page.
+   * @param flagKeys
+   */
+  public applyVariants(flagKeys: string[] | undefined = undefined) {
```
bgiori, line 285: Change to object input, `ApplyVariantsOptions`, add `flagKeys` as optional field.

packages/experiment-tag/src/experiment.ts
bgiori, line 329: Change to object input, `RevertVariantsOptions`, add `flagKeys` as optional field.

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
-      const isControlPayload =
-        !variant.payload || (payloadIsArray && variant.payload.length === 0);
-      if (shouldTrackExposure && isControlPayload) {
-        exposureWithDedupe(key, variant);
+
+  /**
+   * Get redirect URLs for flags.
+   * @param flagKeys
+   */
+  public getRedirectUrls(
```
bgiori, line 345: What's the use case for this function?

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
-const handleRedirect = (action, key: string, variant: Variant) => {
-  const globalScope = getGlobalScope();
-  if (!globalScope) {
-    return;
+  /**
+   * Preview the effect of a variant on the page.
+   * @param key
+   * @param variant
+   */
+  public previewVariant(key: string, variant: string) {
```
bgiori, line 384: Support previewing multiple flags/variants. Maybe we just merge this with applyVariants with another config like `preview`?

packages/experiment-tag/src/experiment.ts
```diff
@@ -8,443 +8,675 @@ import {
...
 
-  const targetUrl = concatenateQueryParamsOf(
-    globalScope.location.href,
-    redirectUrl,
-  );
+  /**
+   * Set URL change listener to revert mutations and apply variants on back/forward navigation.
+   * @param flagKeys
+   */
+  public setUrlChangeListener(flagKeys: string[]) {
```
bgiori, line 465: Is this supposed to have a listener argument?

bgiori (changes requested):
Initial review, will review more soon.

packages/experiment-tag/src/experiment.ts
tyiuhc: This is for the case where the developer want to implement their own redirect handler.

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,654 @@ import {
...
-  } else if (user.web_exp_id && user.device_id) {
-    delete user.device_id;
-    globalScope.localStorage.setItem(
-      experimentStorageName,
-      JSON.stringify(user),
-    );
-  }
+    if (this.globalScope?.webExperiment) {
+      return;
+    }
```
bgiori, line 79: Since we're in a constructor, returning early will still return a constructed but uninitialized instance.

IMO the constructor should always return a valid object. it's up to the caller (in this case, us in script.ts) to manage the instance correctly on the global `webExperiment` object.

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,654 @@ import {
...
+    if (urlParams['VISUAL_EDITOR']) {
+      WindowMessenger.setup();
+      this.globalScope.history.replaceState(
         {},
         '',
-        removeQueryParams(globalScope.location.href, ['PREVIEW', key]),
+        removeQueryParams(this.globalScope.location.href, ['VISUAL_EDITOR']),
       );
+      return;
+    }
```
bgiori, line 94: I think I mislead you when we met about this. We should have most of this in the start() function. 

The constructor should just initialize all the required member fields. I.e. we should not be returning early here.

bgiori (changes requested):

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
+
+    // If no integration has been set, use an amplitude integration.
+    if (!this.globalScope.experimentIntegration) {
+      const connector = AnalyticsConnector.getInstance('$default_instance');
+      this.globalScope.experimentIntegration = new AmplitudeIntegrationPlugin(
+        this.apiKey,
+        connector,
+        0,
+      );
+    }
```
bgiori, line 244: This looks to be a duplicate of the code above on line 223, remove

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
-    );
-  } else if (user.web_exp_id && user.device_id) {
-    delete user.device_id;
-    globalScope.localStorage.setItem(
-      experimentStorageName,
-      JSON.stringify(user),
-    );
-  }
+    if (this.globalScope?.webExperiment) {
+      return this.globalScope?.webExperiment;
```
bgiori, line 81: Given people should not be calling this function, we should pull this check outside the constructor

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
-  const urlParams = getUrlParams();
-  // if in visual edit mode, remove the query param
-  if (urlParams['VISUAL_EDITOR']) {
-    globalScope.history.replaceState(
-      {},
-      '',
-      removeQueryParams(globalScope.location.href, ['VISUAL_EDITOR']),
-    );
-    return;
+    this.globalScope.webExperiment = this;
```
bgiori, line 84: similarly, set the global webExperiment outside constructor:
```
if (globalScope.webExperiment) return globalScope.webExperiment;
globalScope.webExperiment = new DefaultWebExperimentClient();
```

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
+    await this.fetchRemoteFlags();
+    // apply remote variants - if fetch is unsuccessful, fallback order: 1. localStorage flags, 2. initial flags
+    this.applyVariants({ flagKeys: this.remoteFlagKeys });
   }
 
-  // If no integration has been set, use an amplitude integration.
-  if (!globalScope.experimentIntegration) {
-    const connector = AnalyticsConnector.getInstance('$default_instance');
-    globalScope.experimentIntegration = new AmplitudeIntegrationPlugin(
+  static async create(
```
bgiori, line 270: This is where we should handle the globalScope.webExperiment singleton instance

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
+   * Apply evaluated variants to the page.
+   * @param options
+   */
+  public applyVariants(options?: ApplyVariantsOptions) {
+    const { flagKeys } = options || {};
+    const variants = this.getVariants();
+    if (Object.keys(variants).length === 0) {
+      return;
+    }
+    if (!this.globalScope) {
```
bgiori, line 301: globalScope should always be defined.

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
 
-  const targetUrl = concatenateQueryParamsOf(
-    globalScope.location.href,
-    redirectUrl,
-  );
+  /**
+   * Get all variants for the current web experiment context.
+   */
+  public getVariants(): Variants {
+    if (!this.experimentClient) return {};
```
bgiori, line 388: experimentClient should always be defined.

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
-  // apply local variants
-  applyVariants(globalScope.webExperiment.all(), localFlagKeys);
-
-  if (!isRemoteBlocking) {
-    // Remove anti-flicker css if remote flags are not blocking
-    globalScope.document.getElementById?.('amp-exp-css')?.remove();
+  /**
+   * Get the underlying {@link ExperimentClient} instance.
+   */
+  public getExperimentClient(): ExperimentClient | undefined {
```
bgiori, line 287: experimentClient should always be defined.

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,660 @@ import {
...
-  let script: HTMLScriptElement | undefined;
-  if (rawJs) {
-    script = globalScope.document.createElement('script');
-    if (script) {
-      script.innerHTML = `function ${id}(html, utils, id){${rawJs}};`;
-      script.id = `js-${id}`;
-      globalScope.document.head.appendChild(script);
+
+  private handleRedirect(action, key: string, variant: Variant) {
+    if (!this.globalScope) {
```
bgiori, line 448: globalScope should always be defined.

bgiori (commented):
Overall looking good. some changes, mostly around the singleton access and removing code where we're checking if globalScope or experimentClient are defined, even though they should always be defined in this case.

packages/experiment-tag/src/config.ts
```diff
@@ -0,0 +1,20 @@
...
+   * Determines whether the default implementation for handling navigation {@link setDefaultUrlChangeHandler} will be used
+   * If this is set to false, for single-page applications:
+   * 1. The variant actions applied will be based on the context (user, page URL) when the web experiment script was loaded
+   * 2. Custom handling of navigation should be implemented such that variant actions applied on the site reflect the latest context
+   */
+  useDefaultNavigationHandler?: boolean;
+  /**
+   * Determines whether anti-flicker CSS should be applied for experiments with remote properties.
+   */
+  applyRemoteExperimentAntiFlicker?: boolean;
```
bgiori, line 14: remove, per our discussion

packages/experiment-tag/src/experiment.ts
```diff
@@ -4,461 +4,641 @@ import {
...
+      return globalScope.webExperiment;
+    }
+    const webExperiment = new DefaultWebExperimentClient(
       apiKey,
-      connector,
-      0,
+      initialFlags,
+      config,
     );
+    await webExperiment.start();
```
bgiori, line 283: pull the call to `start` out of `getInstance` and make the function synchronous then the caller would need to call start as needed.

also make `start` idempotent (can only be called once per instance).

bgiori (changes requested):

PR #166, opened 2025-03-19
fix: Web experiment action deduplication

tyiuhc:
### Checklist

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -507,10 +507,6 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
   }
 
   private handleInject(action, key: string, variant: Variant) {
-    // Check for repeat invocations
-    if (this.appliedMutations[key]?.[INJECT_ACTION]) {
-      return;
-    }
     // Validate and transform ID
     let id = action.data.id;
     if (!id || typeof id !== 'string' || id.length === 0) {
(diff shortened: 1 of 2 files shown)


packages/experiment-tag/src/experiment.ts
```diff
@@ -474,21 +491,24 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
...
+    (this.appliedMutations[key] ??= {})[MUTATE_ACTION] = mutationControllers;
     this.exposureWithDedupe(key, variant);
   }
 
   private handleInject(action, key: string, variant: Variant) {
     // Check for repeat invocations
-    if (this.appliedMutations[key]) {
+    if (this.appliedMutations[key]?.[INJECT_ACTION]) {
       return;
     }
```
bgiori, line 513: Do we even need this? I feel like the We already check for applied injections at line 522

packages/experiment-tag/src/experiment.ts
```diff
@@ -47,7 +50,11 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
   private readonly globalScope: typeof globalThis;
   private readonly experimentClient: ExperimentClient;
   private appliedInjections: Set<string> = new Set();
-  private appliedMutations: Record<string, MutationController[]> = {};
+  private appliedMutations: {
+    [experiment: string]: {
+      [actionType: string]: MutationController[];
+    };
+  } = {};
```
bgiori, line 57: If we use appliedInjections to track injections and mutations to track mutations, I dont think we need to nest action type in appliedMutations?

tyiuhc: This is required for `revertVariants()` which iterates through  `appliedMutations`

bgiori (approved):

PR #168, opened 2025-03-24
feat: experiment plugin

Mercy811:
### Summary

This PR depends on 
- https://github.com/amplitude/Amplitude-TypeScript/pull/1013
- https://github.com/amplitude/Amplitude-TypeScript/pull/1033

- Use module augmentation to add experiment variable at compile time.
- Use Object.defineProperty to add experiment getter for runtime.  

```ts
import * as amplitude from '@amplitude/analytics-browser';
// amplitude here refers to the module rather than an object.

// Object destructuring is used to assign init as the init of the default instance. 
amplitude.init('API_KEY', 'user@company.com');
amplitude.identify(new amplitude.Identify().set('premium', true));
// Ideally, we want to the following to work, 
// because it requires minimum changes.
// BUT, it's NOT supported.
amplitude.new ExperimentAnalyticsPlugin());
amplitude.experiment.fetch();

// To be able to get access to experiment,
// have to hold a reference to amplitude analytics instance.
const client: AmplitudeBrowser = new amplitude.AmplitudeBrowser();
client.init('API_KEY', 'user@company.com', {
  logLevel: amplitude.Types.LogLevel.Debug,
}).promise.then((_) => {
  client.add(new ExperimentAnalyticsPlugin({ debug: true })).promise.then((_) => {
    // Add experiment plugin to existing Amplitude analytics SDK
    // and call experiment APIs.
    // Experiment APIs are only available when
    // 1. Amplitude instance has successfully setup
    // 2. Experiment plugin has been installed successfully
    void client.experiment.fetch();
    const variant = client.experiment.variant('FLAG_KEY');
  });
});

```

### Checklist

### DIFF
--- a/packages/experiment-browser/src/experiment-analytics-plugin.ts
+++ b/packages/experiment-browser/src/experiment-analytics-plugin.ts
@@ -1,21 +1,26 @@
-import { AmplitudeBrowser, Types } from '@amplitude/analytics-browser';
+import {
+  AmplitudeCore,
+  EnrichmentPlugin,
+  CoreClient,
+  IConfig,
+} from '@amplitude/analytics-core';
 
 import { ExperimentClient } from './experimentClient';
 import { initializeWithAmplitudeAnalytics } from './factory';
 
 // Module augmentation to add experiment variable at compile time
-declare module '@amplitude/analytics-browser' {
-  interface AmplitudeBrowser {
+declare module '@amplitude/analytics-core' {
+  interface CoreClient {
     experiment: ExperimentClient;
   }
 }
 
-export class ExperimentAnalyticsPlugin implements Types.EnrichmentPlugin {
+export class ExperimentAnalyticsPlugin implements EnrichmentPlugin {
   name = 'experiment-analytics-plugin';
   experiment: ExperimentClient;
 
   constructor() {
-    Object.defineProperty(AmplitudeBrowser.prototype, 'experiment', {
+    Object.defineProperty(AmplitudeCore.prototype, 'experiment', {
       get: function () {
         return (
           this.plugin(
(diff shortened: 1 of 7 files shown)


packages/experiment-browser/src/experiment-analytics-plugin.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export class ExperimentAnalyticsPlugin implements Types.EnrichmentPlugin {
+  name = 'experiment-analytics-plugin';
+  experiment: ExperimentClient;
+
+  constructor() {
+    Object.defineProperty(AmplitudeBrowser.prototype, 'experiment', {
+      get: function () {
+        return this.plugin(ExperimentAnalyticsPlugin.name);
```
crleona, line 20: any reason to return the plugin vs the plugin's instance of experiment?

crleona (approved):

packages/experiment-browser/src/experiment-analytics-plugin.ts
Mercy811: I fixed it at https://github.com/amplitude/experiment-js-client/pull/168/commits/dee63ff78ef00db5c6ab8274367db1be1d7af016

daniel-graham-amplitude (approved):
Lgtm, but I don't think I'm as equipped to review this as the other reviewers are.

packages/experiment-browser/src/experiment-analytics-plugin.ts
```diff
@@ -0,0 +1,37 @@
...
+          this.plugin(
+            ExperimentAnalyticsPlugin.name,
+          ) as ExperimentAnalyticsPlugin
+        ).experiment;
+      },
+    });
+  }
+
+  async setup(config: IConfig, client: CoreClient) {
+    this.experiment = initializeWithAmplitudeAnalytics(config.apiKey);
```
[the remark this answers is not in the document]
Mercy811: Thanks Brian. I updated the plugin constructor to accept config.

Mercy811:
Let me create another package for this plugin rather than export it in experiment sdk package directly.

PR #193, opened 2025-06-30
fix: redirect action impression events fired on new page

tyiuhc:
### Checklist

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -812,7 +812,7 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
         JSON.stringify(storedRedirects),
       );
     } catch (error) {
-      console.warn('Error storing redirect information:', error);
+      console.warn('Error storing redirect events:', error);
     }
   }
 
@@ -836,7 +836,7 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
         this.globalScope.sessionStorage.removeItem(redirectStorageKey);
       }
     } catch (error) {
-      console.warn('Error processing stored redirects:', error);
+      console.warn('Error processing stored redirects events:', error);
     }
   }
 }
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -798,7 +798,7 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
   }
 
   private storeRedirectImpressions(flagKey: string, variant: Variant) {
-    const redirectStorageKey = `AMP_EXP_REDIRECT_${this.apiKey.slice(0, 10)}`;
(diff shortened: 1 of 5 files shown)


packages/experiment-tag/src/experiment.ts
```diff
@@ -777,4 +780,46 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
...
+    try {
+      const storedRedirects = JSON.parse(
+        this.globalScope.sessionStorage.getItem(redirectStorageKey) || '{}',
+      );
+
+      // If we have stored redirects, track exposures for them
+      if (Object.keys(storedRedirects).length > 0) {
+        for (const storedFlagKey in storedRedirects) {
+          const storedVariant = storedRedirects[storedFlagKey];
+          this.exposureWithDedupe(storedFlagKey, storedVariant);
```
tyiuhc, line 815: edge case: what if user evaluation results on the redirection is different? E.g. `treatment` on original url, `control` on redirected url. With this logic, `control` would be fired on the redirected url

bgiori: Good call out. I would very much prefer being able to track the exact exposure from the original page. We may need to store the entire exposure in that case. This is the problem with wrapping all the logic from the feature exp sdk.

bgiori: I think the async track function may be the better approach actually

packages/experiment-tag/src/experiment.ts
```diff
@@ -346,6 +346,9 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
       this.urlExposureCache = {};
       this.urlExposureCache[currentUrl] = {};
     }
+
+    this.fireStoredRedirectImpressions();
```
zhukaihan, line 350: Should we check if the current page has redirect or not before firing impressions in case of multiple "nested" redirects?

packages/experiment-tag/src/experiment.ts
```diff
@@ -733,7 +738,18 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
...
-      this.experimentClient.exposure(key);
+      if (forceVariant) {
+        const variantAndSource = {
+          variant: variant,
+          source: 'local-evaluation',
+          hasDefaultVariant: false,
+        };
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        this.experimentClient.exposureInternal(key, variantAndSource);
```
zhukaihan, line 749: Probably want to make this `exposureInternal` public. I'm not sure if minimized script may enforce private or not.

zhukaihan (approved):

packages/experiment-tag/src/experiment.ts
tyiuhc: Just tested, the minimized script won't enforce private

PR #194, opened 2025-07-02
fix: check current url equal to redirection url before tracking redirect impression

tyiuhc:
### Checklist


bgiori (approved):

bgiori:
I think my only concern is if something strips user properties prior to our script running. Then the impression will never get tracked.

PR #220, opened 2025-10-01
fix: do not treat falsy values like 0 and false as undefined

tkainrad:
### Summary

My earlier fix (#143 ) was not complete. This removes remaining instances of early returns when a property value evaluates to false. This allows to correctly evaluate comparisons like `0==0` or `0<1` or `false=false`.

However, we need to make sure that this isn't expected behavior for some use cases. Atm, comparing the `(none)` value that you can select in the conditions builder to number value `0` or boolean value `false` would evaluate to `true` which changes with this PR. 

### Checklist


bgiori:
> @tkainrad I agree that this is the correct evaluation logic, but just concerned that it may break existing customer implementation. @bgiori any thoughts?

Correctness across platforms trumps correctness for a single platform

bgiori (approved):

PR #222, opened 2025-10-02
feat: Cookie consent exposure handling

tyiuhc:
### Summary

This update will support consent-aware firing of `$impression` events.

Behavior in different states:
1. `GRANTED`: will fire events as normal - immediately when variant actions are applied
2. `REJECTED`: script will not be run
3. `PENDING`: events will be stored in memory, and only fired when consent status is changed to `GRANTED`

`$impression` events are update to contain a timestamp to indicate when the event occurred, which will override the timestamp generated by the `analytics-browser` upon processing. (Relevant [PR1](https://github.com/amplitude/experiment-js-client/pull/223), [PR2](https://github.com/amplitude/Amplitude-TypeScript/pull/1317)). This is necessary for timestamp accuracy for delayed `$impression` firing during the `PENDING` state.

### Checklist

### DIFF
--- a/packages/experiment-tag/src/exposure/consent-aware-exposure-handler.ts
+++ b/packages/experiment-tag/src/exposure/consent-aware-exposure-handler.ts
@@ -57,7 +57,7 @@ export class ConsentAwareExposureHandler {
   ) {
     return (event: ExperimentEvent): boolean => {
       if (event?.eventProperties) {
-        event.eventProperties.time = new Date().getTime();
+        event.eventProperties.time = Date.now();
       }
       try {
         if (this.consentStatus === ConsentStatus.PENDING) {
--- a/packages/experiment-tag/src/types.ts
+++ b/packages/experiment-tag/src/types.ts
@@ -2,7 +2,6 @@ import { EvaluationCondition } from '@amplitude/experiment-core';
 import {
   ExperimentConfig,
   ExperimentUser,
-  Variant,
 } from '@amplitude/experiment-js-client';
 import { ExperimentClient, Variants } from '@amplitude/experiment-js-client';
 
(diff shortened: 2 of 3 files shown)


packages/experiment-tag/src/exposure/consent-aware-exposure-handler.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Create a consent-aware wrapper for the track method
+   */
+  private createConsentAwareTrack(
+    originalTrack: (event: ExperimentEvent) => boolean,
+  ) {
+    return (event: ExperimentEvent): boolean => {
+      if (event?.eventProperties) {
+        event.eventProperties.time = new Date().getTime();
```
stephen-choi-amplitude: Assuming we don't account for timezone, `Date.now()` is preferred.

packages/experiment-tag/src/experiment.ts
```diff
@@ -134,6 +136,10 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
 
     this.storage = new ConsentAwareStorage(this.consentOptions.status);
 
+    this.consentAwareExposureHandler = new ConsentAwareExposureHandler(
+      this.consentOptions.status,
+    );
```
stephen-choi-amplitude: Should this be in line 133 block since `this.consentOptions.status` can be undefined? Similarly, should `consentAwareExposureHandler` be initialized if the user has not opted into use consent aware storage?
Or should we consider `this.consentOptions.status` to be GRANTED if not provided?

packages/experiment-tag/test/consent-aware-exposure-handler.test.ts
stephen-choi-amplitude: can we add a test case for when consent status is null/undefined?

packages/experiment-tag/src/experiment.ts
stephen-choi-amplitude: Oh, I thought this was `this.config.consentOptions` not `this.consentOptions`. I see that the private field is set with the default value.

stephen-choi-amplitude (approved):

PR #223, opened 2025-10-02
fix: attach timestamp to impression events

tyiuhc:
### Summary

This update will attach timestamps to `$impression` events when they are incurred. 

This is necessary for consent-aware exposure tracking, which keeps `$impression` events in memory when consent status is `PENDING`. When status is updated to `GRANTED`, the timestamp is required to properly track when the event was incurred, and not when the event was processed by the analytics SDK.

Relevant `analytics-browser` [PR](https://github.com/amplitude/Amplitude-TypeScript/pull/1317).

### Checklist


Mercy811 (approved):
I've limited knowledge about experiment SDK but it looks good to me as `Exposure` is just event properties and it aligns with the other pr in analytics SDK.

PR #224, opened 2025-10-06
feat: add setTracksAssignment

zhukaihan:
### Summary

Add `setTracksAssignment`. 

### Checklist

* Does your PR have a breaking change?: no

### DIFF
--- a/packages/experiment-tag/rollup.config.js
+++ b/packages/experiment-tag/rollup.config.js
@@ -16,7 +16,9 @@ import * as packageJson from './package.json';
 
 let branchName = '';
 try {
-  const fullBranch = execSync('git rev-parse --abbrev-ref HEAD', { encoding: 'utf8' }).trim();
+  const fullBranch = execSync('git rev-parse --abbrev-ref HEAD', {
+    encoding: 'utf8',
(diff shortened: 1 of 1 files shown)


packages/experiment-browser/src/storage/cookie-storage.ts
```diff
@@ -0,0 +1,17 @@
+import { getGlobalScope } from '@amplitude/experiment-core';
+
+import { Storage } from '../types/storage';
+export class CookieStorage implements Storage {
```
tyiuhc: where is `CookieStorage` used?

zhukaihan: My bad. Old code got pushed.

packages/experiment-browser/src/experimentClient.ts
```diff
@@ -717,6 +730,14 @@ export class ExperimentClient implements Client {
     }
   }
 
+  public async setTracksAssignment(doTrack: boolean): Promise<void> {
```
vaibhav-jain-exp: Might be worth adding a method description here for SDK consumers.

PR #230, opened 2025-10-24
feat: Add support for defining custom headers

dlmr:
Configuration option for defining a custom function that can set custom headers on requests. Useful when using a custom proxy that might have additional requirement.

### Summary

Adds a configuration option for defining a custom function that can set custom headers on requests. Useful when using a custom proxy that might have additional requirements.

### Checklist

### DIFF
--- a/packages/experiment-browser/src/config.ts
+++ b/packages/experiment-browser/src/config.ts
@@ -3,7 +3,7 @@ import { ExperimentAnalyticsProvider } from './types/analytics';
 import { ExposureTrackingProvider } from './types/exposure';
 import { ExperimentUserProvider } from './types/provider';
 import { Source } from './types/source';
-import { HttpClient } from './types/transport';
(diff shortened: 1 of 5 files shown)


zhukaihan:
Thanks for submitting the PR! 
We do have `httpClient` option in the config to provide a layer of abstraction for you to manipulate the request and handle the actual HTTP request. We suggest to use the `httpClient` for this use case and you can add in all the headers you want there.

zhukaihan:
Yeah. Please create PR and add the headers for iOS and Android like this one. We'll approve for iOS and Android. Adding a custom http client handler for these SDKs are a bit too much right now. But for JS, please use httpClient. 

Also nit: for doFlags, we should just pass null to the `customRequestHeaders`. There should be no user in request header for flags requests (might be a mistake on our end but no correctness issue there).

PR #234, opened 2025-10-30
feat: Support custom headers for fetch requests

tyiuhc:
### Summary

Configuration option for defining a custom function that can set custom headers on requests.

This update is meant to be a more generic implementation of [this](https://github.com/amplitude/experiment-js-client/pull/230), which supports specifically adding user context.

### Checklist

### DIFF
--- a/packages/experiment-browser/src/config.ts
+++ b/packages/experiment-browser/src/config.ts
@@ -3,7 +3,7 @@ import { ExperimentAnalyticsProvider } from './types/analytics';
 import { ExposureTrackingProvider } from './types/exposure';
 import { ExperimentUserProvider } from './types/provider';
 import { Source } from './types/source';
-import { HttpClient } from './types/transport';
+import { CustomRequestHeaders, HttpClient } from './types/transport';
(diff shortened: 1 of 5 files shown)


packages/experiment-browser/src/config.ts
```diff
@@ -157,6 +157,11 @@ export interface ExperimentConfig {
    * (Advanced) Use your own http client.
    */
   httpClient?: HttpClient;
+
+  /**
+   * (Advanced) Use custom request headers.
+   */
+  customRequestHeaders?: CustomRequestHeaders;
```
zhukaihan: Why they don't use httpClient for manipulating requests before sending off? I feel like we're make everything configurable with httpClient: headers, URL, URL params, even HTTP method.

dlmr: As mentioned in the other PR, it would be great with a unified API for all clients and while the JS client does support a custom httpClient it's not possible in for example the Swift and Kotlin SDKs.

https://github.com/amplitude/experiment-js-client/pull/230#issuecomment-3472143524

PR #236, opened 2025-11-04
feat: allow configs to take in user provided logger instances for client logging behavior 

kyeh-amp:
### Summary

These changes provide users with more controls over the SDK logging behavior. Users can now provide their own logger instances via client configs.

### Checklist

### DIFF
--- a/packages/experiment-browser/src/experimentClient.ts
+++ b/packages/experiment-browser/src/experimentClient.ts
@@ -134,7 +134,7 @@ export class ExperimentClient implements Client {
     };
     this.logger = new AmpLogger(
       this.config.loggerProvider || new ConsoleLogger(),
-      this.getLogLevel(config),
+      ExperimentClient.getLogLevel(config),
     );
     const internalInstanceName = this.config?.['internalInstanceNameSuffix'];
     this.isWebExperiment = internalInstanceName === 'web';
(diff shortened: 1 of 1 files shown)


packages/experiment-browser/src/experimentClient.ts
```diff
@@ -935,11 +941,13 @@ export class ExperimentClient implements Client {
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private debug(message?: any, ...optionalParams: any[]): void {
-    if (this.config.debug) {
-      console.debug(message, ...optionalParams);
+  private getLogLevel(config: ExperimentConfig): LogLevel {
```
zhukaihan: nnnit: Can be static func.

packages/experiment-browser/src/logger/ampLogger.ts
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable @typescript-eslint/no-explicit-any*/
+import { Logger, LogLevel } from '../types/logger';
+
+/**
+ * Internal logger class that wraps a Logger implementation and handles log level filtering.
+ * This class provides a centralized logging mechanism for the Experiment client.
+ * @category Logging
+ */
+export class AmpLogger implements Logger {
```
zhukaihan: I also wonder the necessity of the wrapper given we're not doing any filtering or transformation like other SDKs. We usually want to keep `experiment-browser` lean.

kyeh-amp: Great point about lean skds.  I believe this is a low addition way to provide log level filtering on our end as well as have a decent buffer between our log level definitions/usages and others.

PR #238, opened 2025-11-06
feat: support cross-subdomain redirect impression tracking

tyiuhc:
### Summary

Adds opt-in support for cross-subdomain and cross-domain redirect impression tracking without changing default behavior for existing customers. Two new flags on `redirectConfig` control the fallback strategies:

```js
window.experimentConfig = {
  redirectConfig: {
    encodeRedirectInCookie: true,  // cross-subdomain (root-domain cookie)
    encodeRedirectInUrl: true,     // cross-domain / cookie-blocked (URL param)
  }
}
```

By default, redirect impressions continue to use `sessionStorage` only. When either flag is set, impressions are additionally persisted via the configured fallback(s) so they survive the redirect to a different subdomain or domain.

**Fallback strategies:**

- **`encodeRedirectInCookie`** — writes impression data to a short-lived cookie scoped to the root domain (resolved via `getTopLevelDomain()`). Enables tracking across subdomains that share a root domain (e.g. `app.example.com` → `www.example.com`). Cookie is only written when the redirect target hostname differs from the current hostname — same-subdomain redirects use sessionStorage only.
- **`encodeRedirectInUrl`** — base64-encodes impression data into an `AMP_REDIRECT` query parameter appended to the redirect URL. Enables tracking across domains and in cookie-blocked environments. The param is cleaned from the URL via `history.replaceState` after impressions are fired. URL param is only appended when the redirect target hostname differs from the current hostname.

On the destination page, `fireStoredRedirectImpressions` merges all three sources with priority: **URL > sessionStorage > cookie**.

**Supporting changes:**

- **`getTopLevelDomain()`** (`util/cookie.ts`): New async utility that probes for the broadest writable cookie domain. Includes a `KNOWN_2LDS` list to skip known public two-level suffixes (e.g. `.co.uk`). Result is module-level cached after the first call. Used to scope both the marketing cookie and redirect impression cookies.
- **Cross-subdomain guard**: `isCrossSubdomain` is computed in `handleRedirect` by comparing `new URL(targetUrl).hostname` with `location.hostname`. Cookie writes and URL param encoding are skipped entirely for same-hostname redirects — only sessionStorage is used.
- **Lean `StoredRedirectImpression` type**: Replaces storing the full `Variant` object with a minimal shape (`{ redirectUrl, variantKey, expKey?, metadata? }`) — only the fields needed to reconstruct the exposure event.
- **`setMarketingCookie`**: Now resolves its own domain internally via `getTopLevelDomain()` — no longer accepts a `domain` parameter.
- **Async redirect handling**: Variant application and `fireStoredRedirectImpressions` are now `async`, ensuring all impression storage writes are fully awaited before `location.replace` fires.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core redirect, exposure deduplication, and startup ordering (anti-flicker, async applyVariants); mis-merged impressions or URL cleanup could cause missed or duplicate exposures, though defaults preserve existing sessionStorage-only behavior.
> 
> **Overview**
> Adds opt-in **`redirectConfig`** on web experiments so redirect **exposure/impression** data can survive navigation beyond **`sessionStorage`**, without changing behavior unless customers enable the flags.
> 
> When **`encodeRedirectInCookie`** is on and the redirect crosses hostnames, impressions are written to a short-lived root-domain cookie (via new **`getTopLevelDomain()`**). When **`encodeRedirectInUrl`** is on, data is merged into a base64 **`AMP_REDIRECT`** query param on the destination URL and stripped with **`replaceState`** after firing (with **`markUrlAsPublished`** to avoid duplicate exposures). On landing, **`fireStoredRedirectImpressions`** merges sources with priority **URL > session > cookie**, uses a lean **`StoredRedirectImpression`** payload instead of full variants, and runs earlier on **`start`** plus **`url_change`** subscriptions.
> 
> **Redirect handling is now async** (`applyVariants`, `previewVariants`, `handleRedirect`): storage and marketing cookie writes are awaited before **`location.replace`**. **`setMarketingCookie`** scopes cookies using the resolved top-level domain. Subscription group callbacks **`await`** variant application. Tests add **`CookieStorage`** mocks and cross-subdomain/cross-domain redirect cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cb4da1b8d6660af42ea568aceff93be70bf797a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/experiment-tag/src/util/url.ts
```diff
@@ -84,3 +84,34 @@ export const isPreviewMode = (): boolean => {
   }
   return false;
 };
+
+/**
+ * Extracts the root domain from a URL and returns it with a leading dot for cookie sharing.
+ */
+export const getCookieDomain = (url: string): string | undefined => {
```
kwliou: This incorrectly returns .co.uk for `https://site.co.uk`

kwliou (changes requested):

packages/experiment-tag/src/util/url.ts
tyiuhc: You are right - we can update `getCookieDomain` to replicate [`getTopLevelDomain`](https://github.com/amplitude/Amplitude-TypeScript/blob/48f38dad155c810c85acf684bbc70e97fbe2ce82/packages/analytics-browser/src/config.ts#L495) from `analytics-browser` instead.

ansonchen2 (approved):

PR #242, opened 2025-12-17
feat: add time_on_page trigger

tyiuhc:
### Summary

Support `time_on_page` page trigger, fired when the end user has been on the page for a specified `durationMs`.

### Checklist

### DIFF
--- a/packages/experiment-tag/src/subscriptions.ts
+++ b/packages/experiment-tag/src/subscriptions.ts
@@ -6,8 +6,6 @@ import {
   ExitIntentPayload,
   MessageBus,
   MessagePayloads,
-  ElementAppearedPayload,
-  ManualTriggerPayload,
   AnalyticsEventPayload,
   MessageType,
   TimeOnPagePayload,


packages/experiment-tag/src/subscriptions.ts
```diff
@@ -413,6 +419,31 @@ export class SubscriptionManager {
...
+    this.timeOnPage = 0;
+    // Publish initial time_on_page event
+    this.messageBus.publish('time_on_page');
+
+    for (const pages of Object.values(this.pageObjects)) {
+      for (const page of Object.values(pages)) {
+        if (page.trigger_type === 'time_on_page') {
+          const triggerValue = page.trigger_value as TimeOnPageTriggerValue;
+          const durationMs = triggerValue.durationMs;
+          const timeoutId = setTimeout(() => {
```
stephen-choi-amplitude: 1. It's missing [`visibilitychange` event](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) handler to reset when the user moves away from the current tab
2. Q: When there are multiple page objects with the time on page trigger type, are there any potential issues?

PR #246, opened 2025-12-18
feat: add exit_intent trigger

tyiuhc:
### Summary

Add `exit_intent` page trigger. This trigger will fire when the upward cursor movement near viewport top is detected, and the user has spent at least `minTimeOnPage` (if specified).

### Checklist

### DIFF
--- a/packages/experiment-tag/src/subscriptions.ts
+++ b/packages/experiment-tag/src/subscriptions.ts
@@ -2,11 +2,10 @@ import { EvaluationEngine } from '@amplitude/experiment-core';
 
 import { DefaultWebExperimentClient, INJECT_ACTION } from './experiment';
 import {
-  MessageBus,
-  MessagePayloads,
-  ElementAppearedPayload,
   ManualTriggerPayload,
   ExitIntentPayload,
+  MessageBus,
+  MessagePayloads,
   MessageType,
 } from './message-bus';
 import { DebouncedMutationManager } from './mutation-manager';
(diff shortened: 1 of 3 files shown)


packages/experiment-tag/src/subscriptions.ts
```diff
@@ -373,6 +378,62 @@ export class SubscriptionManager {
     });
   };
 
+  private setupExitIntentPublisher = () => {
+    // Check if any page objects use exit_intent trigger
+    const hasExitIntentTriggers = Object.values(this.pageObjects).some(
```
kwliou: I'm not familiar with the flow but does this `pageObjects` already filter only the active page objects?

nit: if you're using `pages` multiple times, you could assign it at the top first:
```ts
const pages = Object.values(this.pageObjects).flatMap((pages) => Object.values(pages));
```

packages/experiment-tag/src/subscriptions.ts
```diff
@@ -373,6 +378,62 @@ export class SubscriptionManager {
...
+    let minTimeOnPageMs = 0;
+    for (const pages of Object.values(this.pageObjects)) {
+      for (const page of Object.values(pages)) {
+        if (page.trigger_type === 'exit_intent') {
+          const triggerValue = page.trigger_value as ExitIntentTriggerValue;
+          if (triggerValue.minTimeOnPageMs !== undefined) {
+            minTimeOnPageMs =
+              minTimeOnPageMs === 0
+                ? triggerValue.minTimeOnPageMs
+                : Math.min(minTimeOnPageMs, triggerValue.minTimeOnPageMs);
```
kwliou: it looks like you can shorten this to `minTimeOnPageMs = Math.min(minTimeOnPageMs, triggerValue.minTimeOnPageMs)`

kwliou (changes requested):

PR #304, opened 2026-04-09
fix(experiment-tag): load overlay after shell is built in mobile mode

stephen-choi-amplitude:
## Summary

- In mobile mode, `buildShell` now completes before the overlay loads, eliminating a race condition where DOM restructuring disrupted the overlay's React 18 concurrent render
- `buildShell` returns a `Promise` that resolves after the shell is built; `experiment.ts` awaits it before calling `WindowMessenger.setup()`
- `buildShell` waits for `document.readyState === 'complete'` via `requestAnimationFrame` polling to avoid SSR hydration errors, without relying on `window.addEventListener('load')` which third-party scripts can suppress

Follows up on #299 which removed the `window.load` dependency from `asyncLoadScript`.

## Test plan

- [x] Visual editor desktop-to-mobile switch on SSR site (Next.js) — shell + overlay both render, no hydration errors
- [x] Visual editor desktop-to-mobile switch on non-SSR site — shell + overlay both render
- [x] Visual editor on Rocket Loader site — overlay loads on first attempt
- [x] MPA navigation with visual editor open — overlay reloads via stored session
- [x] Desktop mode visual editor — no regression, overlay loads normally

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -238,11 +238,14 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
 
     // if in visual edit mode, remove the query param
     if (this.isVisualEditorMode) {
-      WindowMessenger.setup();
-
       if (isMobileModeActive()) {
-        buildShell(this.globalScope);
+        // In mobile mode, build the shell first and load the overlay after.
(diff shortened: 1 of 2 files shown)


ansonchen2 (approved):
🙌

PR #313, opened 2026-05-02
fix(experiment-tag): CSP-safe widget CSS injection [WEB-6]

stephen-choi-amplitude:
## Summary

- Replaces production widget CSS injection in `experiment.ts` from `<style>` element creation to constructable stylesheets adopted onto the customer document via `Document.adoptedStyleSheets`.
- Constructable stylesheets aren't gated by CSP `style-src`, so this fixes production widget rendering on customer pages with strict nonce/hash CSPs.
- Adds a duplicate of the `cspSafeStyleSheet` helper from the visual editor side (cross-reference comment in source ties the two copies together).
- Adds `construct-style-sheets-polyfill` as a devDependency for jest's jsdom 20 environment (jsdom 20 doesn't natively support constructable stylesheets).

## Companion changes

The visual editor side of this fix lives in an internal Amplitude repo. The same constructable-stylesheet swap is applied there to the editor's widget rendering paths. The two PRs together address the bug end-to-end: the internal one fixes the editor experience, this one fixes production widget rendering on the same customer pages.

## Tickets

- Internal tracking: WEB-6 (production SDK side, this PR).

## Backwards compatibility

- **Variant payload format unchanged.** Both old and new SDK read the same `action.data.css` string from the variant payload. No schema or format change.
- **Old widgets, new SDK:** the CSS string format is identical to what the old SDK consumed; constructable stylesheet adoption applies it to the same selector matches as the old `<style>`-element approach.
- **Edge case (rare):** constructable stylesheets reject `@import` rules per the CSSOM spec; the polyfill silently strips them. Customer widget CSS that uses `@import url('...')` would lose those imports. Real browsers behave the same way. Affected widgets would need to either inline the imported styles or use `<link rel=\"stylesheet\">` separately.
- **Edge case (very rare):** the previous code set `style.id = \"css-\${id}\"`. After this change, no `<style>` element exists in the customer DOM with that id. If any customer custom-JS in their inject action depended on `document.querySelector('#css-\${id}')` to read or mutate the widget's CSS (atypical pattern), that lookup will return null. Out-of-the-box widget templates don't do this.

## Test plan

- [x] Unit tests for `cspSafeStyleSheet` helper (7 new tests, all pass).
- [x] All existing experiment-tag tests pass (148/148).
- [x] Typecheck clean.
- [x] Lint clean (60 pre-existing warnings, 0 new).
- [x] SDK build succeeds.
- [ ] Post-merge: smoke-test a customer page with strict CSP and a published widget, confirm it renders styled (requires the new SDK release to ship and the customer's page to load it via dynamic script).

## Out of scope (intentionally)

- The 3 other `<style>` injection sites in this SDK (`anti-flicker.ts`, `preview/preview.ts`, `util/shell.ts`) — same issue, mechanical follow-up tracked internally as WEB-7.
- Older-browser fallback for browsers lacking `Document.adoptedStyleSheets`. Native support: Chrome 99+, Firefox 101+, Safari 16.4+ — three years of support across all modern browsers; SDK targets es2015 syntactically but customers running 4+ year old browsers are vanishingly rare.
- Squelching pre-existing `construct-style-sheets-polyfill` MutationObserver stderr noise during jsdom test teardown (cosmetic; doesn't fail tests).

## Release coordination

This change requires a new minor version of `@amplitude/experiment-tag` to ship to customer pages. The SDK is delivered via dynamic script with no customer pinning, so the fix propagates to customers on their next page load after the new version is published and CDN propagates (typically minutes).

---

> [!NOTE]
> **Medium Risk**
> Changes production widget CSS injection from `<style>` tags to `adoptedStyleSheets`, which can affect how styles are applied/removed on customer pages (and depends on constructable stylesheet support). Scope is limited to inject-action CSS handling and is covered by new unit tests.
> 
> **Overview**
> Fixes widget rendering on strict CSP pages by switching inject-action CSS application from creating a `<style>` element to adopting a constructable stylesheet via `Document/ShadowRoot.adoptedStyleSheets`.
> 
> Adds a new `cspSafeStyleSheet` helper (with revert/reapply handles and iframe-realm-safe construction), wires it into `handleInject` cleanup, and updates Jest to load `construct-style-sheets-polyfill` with new unit tests for the helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e70fa58a3aaa48205ee81547a3e9070cd1c886e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
+++ b/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
@@ -21,7 +21,7 @@ export type StyleSheetHandle = {
  * Document constructor.
  */
 function isDocument(node: Document | ShadowRoot): node is Document {
-  return node.nodeType === /* Node.DOCUMENT_NODE */ 9;
+  return node.nodeType === Node.DOCUMENT_NODE;
 }
 
 export function cspSafeStyleSheet(
--- a/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
+++ b/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
@@ -18,7 +18,13 @@ export function cspSafeStyleSheet(
   target: Document | ShadowRoot,
   css: string,
 ): StyleSheetHandle {
-  const sheet = new CSSStyleSheet();
+  // CSSStyleSheets are realm-bound — adopting one cross-realm throws
+  // NotAllowedError. Construct the sheet via the target's owning realm so it
+  // can be adopted into an iframe's contentDocument.
+  const ownerDoc: Document =
+    (target as Document).ownerDocument ?? (target as Document);
+  const SheetCtor = ownerDoc.defaultView?.CSSStyleSheet ?? CSSStyleSheet;
+  const sheet = new SheetCtor();
   sheet.replaceSync(css);
   let adopted = false;
 
--- a/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
+++ b/packages/experiment-tag/src/util/csp-safe-stylesheet.ts
@@ -1,0 +1,48 @@
+/**
+ * Adopts a CSSStyleSheet onto a Document or ShadowRoot.
+ *
+ * Constructable stylesheets are not gated by CSP `style-src`, which means this
+ * works on customer pages with strict nonce/hash CSP policies (e.g. Fathom)
+ * where `<style>` element injection would be blocked.
+ *
+ * Returns idempotent revert/reapply handles for hide-and-restore use cases
+ * (e.g. AI stylizer page snapshots).
+ *
+ * NOTE: This file is duplicated in:
+ *   - apps/experiment-overlay/src/lib/csp-safe-stylesheet.ts (javascript repo)
+ *   - packages/experiment-tag/src/util/csp-safe-stylesheet.ts (experiment-js-client repo)
+ * If you change one, change the other.
+ */
+
+export type StyleSheetHandle = {
+  revert: () => void;
+  reapply: () => void;
+};
+
+export function cspSafeStyleSheet(
+  target: Document | ShadowRoot,
+  css: string,
+): StyleSheetHandle {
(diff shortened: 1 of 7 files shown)


packages/experiment-tag/src/util/csp-safe-stylesheet.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Adopts a CSSStyleSheet onto a Document or ShadowRoot.
+ *
+ * Constructable stylesheets are not gated by CSP `style-src`, which means this
+ * works on customer pages with strict nonce/hash CSP policies (e.g. Fathom)
```
kwliou: nit: remove Fathom mention since this is a public repo

stephen-choi-amplitude: Done — also scrubbed the other internal reference in the same JSDoc and dropped the cross-repo NOTE block. Fixed in [`d4919c2`](https://github.com/amplitude/experiment-js-client/commit/d4919c26535bb2b9c51db4f22b314c15c3f55f47).

packages/experiment-tag/src/util/csp-safe-stylesheet.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Discriminates Document from ShadowRoot via nodeType. We avoid
+ * `instanceof Document` because `instanceof` lies across realms — an iframe's
+ * contentDocument IS a Document but isn't an instance of THIS realm's
+ * Document constructor.
+ */
+function isDocument(node: Document | ShadowRoot): node is Document {
+  return node.nodeType === /* Node.DOCUMENT_NODE */ 9;
```
kwliou: You can just reference `Node.DOCUMENT_NODE` directly.

kwliou (approved):

PR #324, opened 2026-05-19
feat: support Mutex and Holdout groups for web experiment

tyiuhc:
### Summary

Support Mutex and Holdout groups for web experiment.

Flags that transitively depend on a remote-evaluated flag (e.g. a Mutex or Holdout parent) are now promoted to remote evaluation via a fixed-point dependency walk. This ensures child flags wait for their remote dependencies to be fetched before bucketing runs, preventing incorrect variant assignment.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes flag evaluation ordering and startup timing for experiments with flag dependencies or holdout/mutex flags, which can affect which variants users see on first paint.
> 
> **Overview**
> Web experiment startup now treats **local flags that depend on a remote flag** (including transitive chains) as remote so variant application waits until remote parents are fetched, avoiding wrong mutex/holdout bucketing from stale parent state. Only flags promoted via that dependency walk are removed from the early **local** apply pass; directly remote flags that are session-cached as local still get the fast-path local apply.
> 
> When any initial flag key starts with **`holdout-`** or **`mutex-`**, **`start()`** awaits **`integrationManager.ready()`** before applying variants so bucketing has **`user_id` / `device_id`** from the Amplitude integration setup.
> 
> Adds a test that a depth-2 local→local→remote dependency chain is not applied on the first local **`applyVariants`** call and is applied after remote fetch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d232bbfaa35c17d6f939f41e06942596b29c129a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -413,6 +413,12 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
     this.experimentClient.setUser(enrichedUser);
     this.updateUserWithBehaviors();
 
+    // Wait for the integration's setup() to complete before evaluating local
+    // flags — otherwise bucketing runs against incomplete user identity.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    await this.experimentClient.integrationManager.ready();
+
     if (!this.isRemoteBlocking) {
       removeAntiFlickerCss();
     }
--- a/packages/experiment-tag/src/experiment.ts
+++ b/packages/experiment-tag/src/experiment.ts
@@ -182,6 +182,22 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
       }
     });
 
+    // Flags that depend on at least one remote flag are also considered remote.
+    // Iterate to fixed point to handle transitive dependencies.
+    let changed = true;
+    while (changed) {
+      changed = false;
+      this.initialFlags.forEach((flag: EvaluationFlag) => {
+        if (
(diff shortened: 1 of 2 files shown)


ansonchen2 (approved):

packages/experiment-tag/src/experiment.ts
```diff
@@ -182,6 +182,27 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
...
     });
 
+    // Flags that depend on at least one remote flag are also considered remote.
+    // Iterate to fixed point to handle transitive dependencies.
+    // Track which keys were promoted so we can exclude only those from
+    // localFlagKeys — flags that are directly 'remote' but cached in session
+    // storage as 'local' should still be applied locally (fast-path rendering).
+    const transitivelyPromotedKeys = new Set<string>();
+    let changed = true;
+    while (changed) {
```
stephen-choi-amplitude, line 196: Correct me if I'm wrong, but the outer while loop only matters for chains of depth 2+. The test only covers depth 1, so it would pass identically with a single forEach pass.

Could we either add a transitive test (e.g. C → B → A with initialFlags ordered [C, B, A]) to justify the fixed-point, or drop it for a single pass if mutex/holdout depth-1 is all we need?

tyiuhc: Updated unit test to test for 2+ depth

PR #335, opened 2026-06-16
feat(experiment-tag): Wire relay iframe on start

tyiuhc:
## Summary

Stacks on [#333](https://github.com/amplitude/experiment-js-client/pull/333) (`web/rtbt-relay-client`). Injects the CDN relay iframe from the live SDK `start()` path when RTBT rules are present.

- **`experiment.ts`**: `scheduleRelaySync()` after Pass 1 `applyVariants` (non-blocking); creates `RelayClient(apiKey, webExpIdV2, getRelayUrl(apiKey, config.serverZone))` so EU deployments load the relay from `cdn.eu.amplitude.com`; Pass 2 re-apply via `handleRelayPass2` when behaviors change
- **`BehavioralTargetingManager`**: `beginRelaySync()` — `relay.init()`, attach client, run `syncFromRelay` when [#334](https://github.com/amplitude/experiment-js-client/pull/334) storage hooks land
- **`WebExperimentUser`**: optional `web_exp_id_v2` field (uses `web_exp_id` fallback until #332 cookie resolve lands)

[#334](https://github.com/amplitude/experiment-js-client/pull/334) (`web/rtbt-use-relay`) stacks on this branch for event storage dual-write.

## Test plan

- [x] `npx jest test/behavioral-targeting/behavioral-targeting-manager-relay.test.ts`
- [x] `npx jest test/experiment-relay-iframe.test.ts`
- [x] `npx jest test/behavioral-targeting/relay-client.test.ts` (base #333)

## Merge order

1. #332 (identity cookies) + nova#31098
2. #333 (`web/rtbt-relay-client` — RelayClient foundation)
3. **This PR** (#335 — iframe wiring)
4. #334 (`web/rtbt-use-relay` — event storage)
5. nova#31184 + deploy#23235 (relay hosting)

---

> [!NOTE]
> **Medium Risk**
> Adds a second startup phase that can re-fetch remote flags and re-apply variants after relay merge, affecting experiment assignment timing on multi-subdomain sites; storage merge is still a no-op until the stacked PR lands.
> 
> **Overview**
> Wires **cross-subdomain behavioral targeting (RTBT)** through a CDN **relay iframe** after the normal `start()` variant pass, without blocking first paint.
> 
[...]

PR #334, opened 2026-06-16
feat(experiment-tag): Dual-write RTBT events to relay

tyiuhc:
## Summary

Stacks on [#335](https://github.com/amplitude/experiment-js-client/pull/335) (`web/rtbt-relay-iframe` → [#333](https://github.com/amplitude/experiment-js-client/pull/333)). Implements RTBT event storage dual-write and Pass 2 merge.

- **`EventStorageManager`**: dual-write on `addEvent`, `mergeFromRelay`, `syncFromRelay`, `flushRelay`
- **`BehavioralTargetingManager`**: direct `setRelayClient` / `syncFromRelay` (used by `beginRelaySync` from #335)

Completes the relay path started in #335: iframe init → attach client → merge relay store → re-evaluate.

## Test plan

- [x] `npx jest test/behavioral-targeting/event-storage.test.ts`
- [x] `npx jest test/behavioral-targeting/behavioral-targeting-manager-relay.test.ts`
- [x] `npx jest test/experiment-relay-iframe.test.ts`
- [x] `npx jest test/behavioral-targeting/relay-client.test.ts`

## Merge order

1. #332 (identity) + nova#31098
2. #333 (`web/rtbt-relay-client` — RelayClient foundation)
3. #335 (`web/rtbt-relay-iframe` — iframe wiring)
4. **This PR** (`web/rtbt-use-relay` — event storage)
5. nova#31184 + deploy#23235 (relay hosting)

---

> [!NOTE]
> **Medium Risk**
> Changes behavioral targeting event persistence and cross-subdomain merge; incorrect sync could skew RTBT counts or duplicate/drop events, though failures degrade to `syncFromRelay` returning false.
> 
> **Overview**
> **`EventStorageManager`** now keeps RTBT events in sync with the cross-subdomain relay instead of stubbing relay hooks.
> 
> New events are **dual-written** to the attached `RelayClient` on `addEvent`; unload/hidden/cleanup paths call **`flushRelay`** via `flush()`. **`eventDedupKey`** (`event_type:timestamp:id`) drives merge deduping so same-millisecond events do not collapse.
> 
[...]

PR #333, opened 2026-06-15
feat(experiment-tag): add RTBT relay protocol/RelayClient and consolidate cross-subdomain identity cookie

tyiuhc:
## Summary

This PR has two parts: a **cross-subdomain identity cookie consolidation** (active for all web-experiment visitors once shipped) and the **RTBT relay foundation** (new plumbing, not yet wired into `start()` / event storage).

### Identity cookies (behavior change — affects every web-experiment visitor)

- Consolidate `web_exp_id_v2` and `first_seen` into a single root-domain **`_identity` JSON cookie** via a new `resolveCrossSubdomainObject` util (one read + at most one write), replacing the two per-field `resolveCrossSubdomainValue` calls. The old helper is removed.
- Resolve the cross-subdomain cookie domain (`rootDomain`) **once, early in `start()`** — after the synchronous `url_change` so anti-flicker variants still apply before the first `await` — and reuse it for every `EXP_` cookie write (identity now, RTBT session later).
- `start()` ordering change: domain resolution now runs before the identity / `first_seen` writes.

### RTBT relay foundation (not yet integrated into `experiment.ts` / event storage)

- Add `relay-protocol.ts` with cross-subdomain RTBT message types and `RELAY_RPC_TIMEOUT_MS` (2000ms)
- Add `RelayClient` — iframe postMessage RPC to `https://{CDN}/script/{API_KEY}.relay.html`
- **Server-zone aware relay URL** — `getRelayUrl(apiKey, serverZone)` returns `cdn.eu.amplitude.com` for the EU zone and `cdn.amplitude.com` otherwise (relay hosting is routed on both CDNs via nova#31184 / deploy#23235); `RelayClient` origin checks follow automatically since `relayOrigin` is derived from the relay URL
- **`relayUrl` config override** on `WebExperimentConfig` — overrides the relay iframe origin (keeps the canonical `/script/{apiKey}.relay.html` path) for pointing at a local/staging dynamic-script; defaults to the CDN per server zone when unset
- **Per-visitor relay namespace** — constructor takes `web_exp_id_v2`; every RPC includes it so CDN `localStorage` is `EXP_{apiKeySlice}_{web_exp_id_v2}_*` (not one store per deployment key per browser)
- Harden init lifecycle: concurrent init dedupe, outer timeout, late ready/body, origin + source validation, destroy/re-init
- Extend `whenBodyReady` with a cancel disposer; `RelayClient.destroy()` stops body polls
- Export relay modules from behavioral-targeting index

[...]

PR #332, opened 2026-06-11
feat(experiment-tag): add cross-subdomain web_exp_id_v2 and first_seen cookies

tyiuhc:
### Summary

SDK-side **Plan B** identity for cross-subdomain web experiments. Resolves `web_exp_id_v2` and `first_seen` as root-domain cookies on `start()`, sets both on the user before `getVariants()` / anti-flicker.

**Product scope:** Old experiments are **not** expected to bucket consistently across subdomains (per-origin `web_exp_id` unchanged). This PR enables:
- **New experiments** — shared `web_exp_id_v2` for cross-subdomain bucketing when paired with [nova#31098](https://github.com/amplitude/nova/pull/31098) (new-experiment `createdAt` cutoff)
- **RTBT** — shared `web_exp_id_v2` for relay storage (`EXP_{slice}_{web_exp_id_v2}_*`)

### `web_exp_id_v2` resolution (cookie authoritative)

1. `EXP_{slice}_web_exp_id_v2` cookie exists → use it
2. Else `web_exp_id_v2` in per-origin localStorage → restore cookie
3. Else `web_exp_id` in per-origin localStorage → seed v2 from same value, set cookie
4. Else generate UUID → assign to **both** `web_exp_id` and `web_exp_id_v2`, set cookie

Mirrors resolved values back to per-origin `EXP_{slice}` localStorage for ITP recovery.

### `first_seen`

- Root-domain cookie `EXP_{slice}_first_seen`
- Migrates from `EXP_{slice}_DEFAULT_USER_PROVIDER` localStorage on first load
- Resolved **before** `getVariants()` so anti-flicker uses shared `first_seen`

### Helper

`resolveCrossSubdomainValue()` in `cookie.ts` — cookie → localStorage fallback → generate → set cookie (reused by the #336 session cookie later).

### Related PRs / merge order

1. **This PR** — client identity
2. [nova#31098](https://github.com/amplitude/nova/pull/31098) — skylab bucketing key (`web_exp_id_v2` for new experiments)
3. [#333](https://github.com/amplitude/experiment-js-client/pull/333) → [#335](https://github.com/amplitude/experiment-js-client/pull/335) / [#334](https://github.com/amplitude/experiment-js-client/pull/334) — relay wiring + dual-write
4. [nova#31184](https://github.com/amplitude/nova/pull/31184) — relay hosting

### Test plan

- [x] `npm test -- experiment.test.ts` (experiment-tag) — 36 tests, including seed-from-`web_exp_id`
[...]

PR #336, opened 2026-06-18
feat(experiment-tag): share RTBT session across subdomains with rolling timeout

tyiuhc:
## Summary

Reworks the behavioral-targeting (RTBT) `SessionManager` so that `current_session` conditions work across subdomains, mirroring Amplitude Analytics' session model.

Previously the RTBT session lived in `sessionStorage`, which is scoped per-tab and per-origin. That fragmented sessions across subdomains (and even tabs), so `current_session` targeting could never match events fired on a different subdomain.

### Changes
- **Root-domain cookie storage** — session id + `lastEventTime` are persisted in a cookie scoped to the registrable domain (resolved with a new synchronous `getTopLevelDomainSync` probe in `util/cookie.ts`), so the session is shared across subdomains and tabs.
- **Rolling inactivity timeout** — the session rotates after 30 minutes of inactivity (same default as Analytics), configurable via the new `WebExperimentConfig.rtbtSessionTimeout`.
- **Activity = any Amplitude event** — `recordActivity()` is called at the top of `BehavioralTargetingManager.trackEvent`, upstream of the `persistedEvents` allowlist filter, so any observed event keeps the session alive (not just RTBT-relevant events). Reads use `getCurrentSessionId()` which does not extend the session.
- **Graceful fallback** — when cookie I/O is blocked (private browsing, ITP), detected via a write-back check, the session degrades to an in-memory per-page session instead of throwing.
- `getOrCreateSessionId()` is kept as a backwards-compatible alias.

### Related PRs
Part of the cross-subdomain behavioral-targeting effort: #333 (RTBT relay protocol + RelayClient), #334 (dual-write events to relay), and #335 (wire relay iframe). The root-domain cookie reuses the same registrable-domain approach as #332 (cross-subdomain `web_exp_id_v2` identity). This change is independent of the relay PRs (based on `main`) but pairs with them so a shared `current_session` matches across subdomains.

## Test plan
- [x] `SessionManager` unit tests: cookie persistence, cross-instance sharing, rolling timeout (extend vs rotate), configurable timeout, start-time, clear, and in-memory fallback.
- [x] Updated event-storage / integration / evaluator suites to clear the session cookie between tests.
- [x] Full `experiment-tag` suite green (174 tests), eslint + prettier clean.
- [ ] Manual cross-subdomain verification (set a `current_session` rule, fire events on `a.example.com`, confirm match on `b.example.com`).


packages/experiment-tag/src/behavioral-targeting/event-storage.ts
```diff
@@ -178,12 +199,11 @@ export class EventStorageManager {
       const relayStore = await relay.readEvents();
 
       // Backfill any local events the relay is missing, regardless of migrated
-      // state. On a first migration, MIGRATE_EVENTS dedupes on
```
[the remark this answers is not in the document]
tyiuhc: Good catch. The uuid change has since been split out of this PR into #342, and the fix landed there in 032b82f: `loadFromLocalStorage` now mints a distinct uuid for any record persisted before uuid existed, so legacy records keep separate dedup identities instead of collapsing under an undefined key on the next `mergeFromRelay`. (This branch no longer carries the uuid change.)

packages/experiment-tag/src/util/cookie.ts
```diff
@@ -122,6 +122,54 @@ const KNOWN_2LDS = [
...
+}
+
+/**
+ * Synchronous variant of {@link getTopLevelDomain}. Resolves the registrable
+ * (root) domain for `hostname` so callers can set a cookie shared across
+ * subdomains, without the async `CookieStorage.isDomainWritable` round-trip.
+ * Returns a leading-dot domain (e.g. `.example.com`) or `''` when no
+ * cross-subdomain domain is writable (single-label hosts, IPs, blocked I/O).
+ */
+export function getTopLevelDomainSync(hostname: string): string {
```
stephen-choi-amplitude, line 150: nit: duplicate logics to `getTopLevelDomain`

Also, can you update comment for https://github.com/amplitude/experiment-js-client/blob/86673efa8a0a1db580a0722d839f83c41ee00699/packages/experiment-tag/src/experiment.ts#L391

packages/experiment-tag/src/behavioral-targeting/session-manager.ts
```diff
@@ -1,79 +1,151 @@
...
+
+  private domain(): string {
+    if (this.resolvedDomain === undefined) {
+      this.resolvedDomain =
+        typeof location !== 'undefined' && location.hostname
+          ? getTopLevelDomainSync(location.hostname)
+          : '';
+    }
+    return this.resolvedDomain;
   }
```
stephen-choi-amplitude, line 150: non-blocker, but... do we need two TLD-resolution paths? The async one falls back to `""` cache on failure, so in the worst case they could resolve to different scopes, granted it's a low chance. Since rootDomain is already resolved earlier, could we just pass `() => this.rootDomain` into SessionManager and drop the sync one? wdyt?

```
this.rootDomain = await getTopLevelDomain(hostname);
this.behavioralTargetingManager = new BehavioralTargetingManager(
  this.apiKey,
  this.behavioralTargetingRules,
  this.config.rtbtSessionTimeout,
  this.rootDomain,  // pass it here
);
```

packages/experiment-tag/src/util/cookie.ts
```diff
@@ -194,6 +242,118 @@ export async function resolveCrossSubdomainObject<
...
+      const raw = readRawCookie(this.key);
+      const parsed = raw !== undefined ? this.parse(raw) : undefined;
+      if (parsed !== undefined) return parsed;
+    }
+    return this.memory;
+  }
+
+  write(value: T): void {
+    this.memory = value;
+    this.usable =
```
stephen-choi-amplitude, line 333: qq: Just want to confirm. Does usable ever go back to true?

stephen-choi-amplitude (approved):
a few minor non-blockers but overall  LGTM

packages/experiment-tag/src/util/cookie.ts
tyiuhc: Fixed the stale doc reference in 51304ec — the comment at experiment.ts pointed at a `getResolvedTopLevelDomain()` that no longer exists; it's `getTopLevelDomainSync()` now.

On the duplication: the two resolvers intentionally share only the candidate-domain walk but differ in the writability probe (sync `isDomainWritableSync` vs the async `CookieStorage.isDomainWritable`), since the session cookie needs a domain synchronously at construction while identity resolution can await. Happy to extract the shared `KNOWN_2LDS` walk into a single helper as a follow-up cleanup, but leaving it out of this PR to keep the approved diff behavior-only.

packages/experiment-tag/src/behavioral-targeting/session-manager.ts
tyiuhc: Passing `() => this.rootDomain` isn't safe with the current lifecycle. `SessionManager` is constructed in the `DefaultWebExperimentClient` constructor (via `BehavioralTargetingManager`), and that constructor runs `evaluateAll()` synchronously — which can hit `getCurrentSessionId() → resolve()` and write a fresh session cookie right then. But `this.rootDomain` isn't resolved until the `await getTopLevelDomain(...)` in `start()`, which is later.

So a closure over `rootDomain` would be `''` at that first write → a host-only cookie. Once `start()` resolves `.example.com`, later writes target the domain cookie but the earlier host-only one shadows the read-back, which trips the `usable` latch and permanently disables cross-subdomain sync for the page. The sync resolver exists precisely to give the session cookie the correct registrable domain at construction time.

The divergence you're describing (async caching `''` on probe failure while sync succeeds) is real but low-probability — same algorithm, same 2LD list, just sync vs async probe. If we want one source of truth without the host-only window, the safe shape is the inverse: resolve synchronously in the constructor and feed that single value to both paths. I'd rather do that as a dedicated follow-up with its own ordering tests than fold it into this approved PR. Filing it.

packages/experiment-tag/src/util/cookie.ts
tyiuhc: No — it's a one-way latch by design. Once a read-back fails, `usable` is set to `false` and the `this.usable !== false` guard short-circuits every subsequent `write`, so `writeRawCookie` is never retried and it stays `false` for the page's lifetime. That's the intended behavior for persistent causes (private mode / ITP / blocked cookie I/O), and it's covered by the `SyncJsonCookie` tests in cookie.test.ts (the "stale cookie shadows a dropped write" and blocked-cookie cases).

The one scenario where a permanent latch is arguably too aggressive is a *transient* host-only shadow (e.g. an upgrade leaving a host-only cookie that shadows the first `.domain` write once). That overlaps with the follow-up from the thread above; if we make domain resolution single-source there, that transient case goes away. Leaving the latch as-is here since it's correct for the persistent case.

packages/experiment-tag/src/behavioral-targeting/session-manager.ts
stephen-choi-amplitude: ah I meant passing the resolved value (i.e. `this.rootDomain`), not a closure as shown in the snippet. IOW, move `BehavioralTargetingManager` initialization to `start()`. Does anything actually need `evaluateAll` before `start()`?

PR #311, opened 2026-04-30
fix: harden safeGlobal consumers and SSR location access

kyeh-amp:
## Summary
- Adds internal helpers in `@amplitude/experiment-core` (`getLocalStorage`, `getSessionStorage`, `getDocument`, `getFetch`, `getSetTimeout`/`getClearTimeout`, `getSetInterval`/`getClearInterval`) that gate browser-API access through `getGlobalScope()` plus an explicit property check, returning `undefined` when the global or property is missing.
- Migrates every unguarded `safeGlobal.X` deref in `experiment-browser` and `plugin-segment` to the helpers; module-load global writes (`factory.ts`, `plugin-segment/plugin.ts`, `experiment-tag/experiment.ts`) no-op when no global is present.
- Fixes #253 (iOS Safari `globalThis === null` crash in `PersistentTrackingQueue`) and the Node SSR crash on `globalScope.location.href` in `convertUserToContext`. Supersedes #256, which was a partial fix; this folds in #256's `safeGlobal`/`getGlobalScope` null-handling and extends it to every unguarded site. `safeGlobal` is now typed `typeof globalThis | undefined`.

## Test plan
- [x] 19 unit tests for the new helpers (global undefined / property missing / property present, plus end-to-end timer schedule + cancel)
- [x] Regression test for iOS Safari class — `PersistentTrackingQueue.setTracker`/`.push` and `SessionDedupeCache` constructor + `shouldTrack` no-throw when the global-scope helpers return `undefined`
- [x] Regression test for Node SSR class — `convertUserToContext` does not throw and omits `context.page` when the global has no `.location`; positive case with `.location` still includes `page.url`
- [x] All existing tests pass across `experiment-core`, `experiment-js-client`, `experiment-tag`, `experiment-plugin-segment`, `analytics-connector` (494 total)
- [x] Node smoke test against built `experiment-browser` UMD bundle: `Experiment.initialize` → `setUser` → `addPlugin` → `variant()` does not throw without browser globals

---

> [!NOTE]
> **Medium Risk**
> Touches core initialization, exposure dedupe/queue persistence, HTTP timeouts, and SSR evaluation context across multiple packages; behavior changes when globals or APIs are missing, though intended paths are guarded and well-tested.
> 
> **Overview**
> Replaces direct **`safeGlobal`** property access with **`@amplitude/experiment-core`** helpers that resolve the global scope (including **`null`** **`globalThis`**, e.g. iOS Safari) and return **`undefined`** when a browser API is missing.
> 
[...]

PR #341, opened 2026-07-01
chore(eslint): restrict globalThis/window/self and use getGlobalScope() [WEB-37]

stephen-choi-amplitude:
## Summary

- Add ESLint `no-restricted-globals` for `globalThis`, `window`, and `self` (disabled in `*.test.ts` / `*.spec.ts`).
- Route runtime global access through `getGlobalScope()` from `@amplitude/experiment-core` in `event-storage`, `spa-link-interceptor`, and related call sites.
- Export `GlobalScope` type from experiment-core for type-only annotations; `analytics-connector` keeps a file-level eslint-disable (standalone shim, no experiment-core dep).

## Test plan

- [x] `yarn lint`
- [x] `packages/experiment-core` — `test/util/global.test.ts` (19 passed)
- [x] `packages/experiment-tag` — `test/behavioral-targeting/event-storage.test.ts` (36 passed)

### DIFF
--- a/packages/experiment-core/src/util/global.ts
+++ b/packages/experiment-core/src/util/global.ts
@@ -1,4 +1,8 @@
-export const getGlobalScope = (): typeof globalThis | undefined => {
+// eslint-disable-next-line no-restricted-globals -- canonical type for the runtime global object
+export type GlobalScope = typeof globalThis;
+
+/* eslint-disable no-restricted-globals -- must probe globals to resolve the runtime scope */
+export const getGlobalScope = (): GlobalScope | undefined => {
   if (typeof globalThis !== 'undefined' && globalThis !== null) {
     return globalThis;
   }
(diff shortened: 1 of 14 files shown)


packages/experiment-core/src/util/global.ts
```diff
@@ -1,4 +1,8 @@
-export const getGlobalScope = (): typeof globalThis | undefined => {
+// eslint-disable-next-line no-restricted-globals -- canonical type for the runtime global object
+export type GlobalScope = typeof globalThis;
+
+/* eslint-disable no-restricted-globals -- must probe globals to resolve the runtime scope */
+export const getGlobalScope = (): GlobalScope | undefined => {
```
kwliou, line 5: I don't know why we don't just use `globalThis` directly? `globalThis` has been supported for a while now.

stephen-choi-amplitude: Two things:
1. `globalThis`/`window` derefs crashed in some Safari lifecycles. see #311 
2. same `no-restricted-globals` is applied in `Amplitude-TypeScript`

tyiuhc (approved):

PR #352, opened 2026-07-14
feat(experiment-tag): gate script on cookie consent (v0)

tyiuhc:
## Summary

Opt-in cookie-consent gating for the web experiment script (v0 / MVP). When `consentOptions.consentRequired` is set and consent isn't `granted`, the client is never constructed — no storage (cookies / localStorage / sessionStorage), evaluation, variant application, tracking, or relay. Granting starts it once.

Off by default: without `consentOptions`, behavior is unchanged.

## Behavior

| Status | Result |
| --- | --- |
| `granted` | Runs normally. |
| `pending` (default when required) | Nothing runs; `pending -> granted` starts the client. |
| `denied` | Nothing runs; behaves like `pending` — a later `granted` (`denied -> granted` re-opt-in) starts the client. |

`pending -> granted` (and `denied -> granted`) starts fresh at grant time, so variants may briefly flicker (**accepted in v0**). Mid-session revocation (`granted -> denied`) is "reload to reset" — a running (or in-flight) client is not torn down until the next load.

Status values follow Google Consent Mode vocabulary (`granted` / `denied` / `pending`).

## What changed

- **`types.ts`** — `ConsentStatus`, `ConsentOptions`, `WebExperimentConfig.consentOptions`. Not in `Defaults`, so absence = off.
- **`consent-gate.ts`** (new, internal) — module-scoped gate state (`deferredStart` / `status` / `started`) + test-only `reset()`. Not re-exported from `index.ts`, keeping the test helper off the public API.
- **`index.ts`** — `initialize()` stashes args and returns early when gated; `setConsentStatus('granted')` starts the client once (clearing the pre-grant buffer), including after `denied` (re-opt-in); `pending`/`denied` both defer. `setConsentStatus` is on the pre-init stub and re-attached to the real client, so early/late CMP callbacks both work. Window config wins over the init arg (existing precedence).
- **`test/consent/`** — gate unit tests (all transitions incl. `denied -> granted` re-opt-in, grant-before-init, precedence, stub, preview-mode races) + real-client journey tests (no storage while pending; running + persisted after grant). 23 tests.

## Risk

- **Precedence**: window config overrides the init arg (matches existing `WebExperimentConfig` resolution).
- **Fail-closed**: `consentRequired` with no status defaults to `pending` — a never-resolved banner means no experiments (intended).
- **Grant-before-init**: an early `granted` is recorded and honored on first init.
- **Pre-load race**: `window.webExperiment` exists only after the tag's `initialize()`. Race-free path: set the initial decision declaratively via `window.experimentConfig.consentOptions.consentStatus`; guard runtime CMP calls with optional chaining. (Customer-facing docs to follow separately.)

## Test plan

- [x] `yarn test` (21 consent tests; full suite 259 passing) / `yarn lint` (0 errors, pre-push)
- [ ] Manual: served-build smoke test with a CMP (deferred to javascript-repo Cypress e2e)

<details>
<summary>AI Prompt</summary>

Implement a v0 for web-experiment cookie consent: `pending` fully stops experiments, `pending -> granted` may flicker (`setConsentStatus` -> init + start + applyVariants, no in-memory storage while pending). Include types, gate + setConsentStatus, tests, README. Follow-up: make `rejected` terminal and update docs/tests.

</details>


packages/experiment-tag/src/index.ts
```diff
@@ -23,9 +70,32 @@ export const initialize = (
...
+    if (status === 'rejected') {
+      consentGate.rejected = true; // terminal at load: never start this load
+      consentGate.deferredStart = null;
+      return;
+    }
+    if (status !== 'granted') {
+      consentGate.deferredStart = { apiKey, initConfigs, config };
+      return;
+    }
+    consentGate.started = true;
```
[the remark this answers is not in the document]
tyiuhc: Fixed in d056ecd — initialize() now returns early when the rejected latch is set (even if its own config says granted), and startClient() re-checks the latch so a rejection during the in-flight preview-config fetch can no longer construct the client. Added regression tests for both paths.

packages/experiment-tag/src/index.ts
```diff
@@ -13,6 +19,47 @@ const eventBuffer: Array<{
...
+  if (
+    status === 'granted' &&
+    consentGate.deferredStart &&
+    !consentGate.started
+  ) {
+    consentGate.started = true;
+    const { apiKey, initConfigs, config } = consentGate.deferredStart;
+    consentGate.deferredStart = null;
+    eventBuffer.length = 0; // drop pre-grant buffer so it isn't replayed
+    void startClient(apiKey, initConfigs, config);
```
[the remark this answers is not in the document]
tyiuhc: Fixed in d056ecd — the deferred grant now routes through a shared launchClient() used by initialize(), so preview/extension sessions get the anti-flicker CSS + latest-config fetch on grant instead of starting directly against the embedded configs. Covered by a new preview-mode test.

packages/experiment-tag/src/index.ts
```diff
@@ -13,6 +19,47 @@ const eventBuffer: Array<{
   event_properties?: Record<string, unknown>;
 }> = [];
 
+const resolveConsentOptions = (
+  config: WebExperimentConfig,
+  globalScope: NonNullable<ReturnType<typeof getGlobalScope>>,
+): ConsentOptions => ({
+  ...config.consentOptions,
+  ...globalScope.experimentConfig?.consentOptions, // window wins (existing precedence)
```
kwliou: I noticed this global config isn't really documented anywhere? I only found one reference https://amplitude.com/docs/web-experiment/proxy for flagsServerUrl

packages/experiment-tag/src/index.ts
```diff
@@ -13,6 +19,43 @@ const eventBuffer: Array<{
...
+/**
+ * Updates cookie-consent status. Exposed on `window.webExperiment` (incl. the
+ * pre-init stub) so a CMP callback can call it before the client exists. v0:
+ * `granted` starts the client once; `rejected` is terminal (latches closed until
+ * reload); `pending` is a no-op. A grant recorded before `initialize()` runs is
+ * honored on init.
+ */
+export const setConsentStatus = (status: ConsentStatus): void => {
+  if (consentGate.status === 'rejected') {
+    return; // terminal: ignore later status changes, including 'granted'
```
kwliou: This is an edge case and I'm not sure what it legally means but on a onetrust site I can click Reject in the banner but then click on their preferences button and re-opt in. Vice versa with accepting and then opting out. If it's also our responsibility to handle consent being revoked dynamically, then we should handle this.

tyiuhc: Good catch — splitting this into its two directions:

`denied → granted` (re-opt-in): agree, fixing in this PR. CMPs treat consent as a stream of updates (OneTrust re-fires OptanonWrapper on every preference-center change; Google Consent Mode and TCF work the same way), and a re-opt-in is just valid new consent. Since nothing is persisted while denied, starting fresh on a later grant is clean. The terminal latch was stricter than necessary — I'll drop it so a later granted starts the client.

`granted → denied` (revocation): keeping "reload to reset" for v0. Honoring revocation on a running client means stop-tracking + storage cleanup — that's the v1 work (stop tracking, delete all EXP_* data in-session). For reference, none of the major experimentation vendors delete storage on revocation: Optimizely writes an opt-out cookie and stops (re-grant often needs a reload), VWO stops immediately but leaves existing cookies and only honors decline-then-accept from the next visit, and Kameleoon sidesteps it by never writing pre-consent. Our own Analytics SDK (setOptOut) is forward-only too. In the meantime, customers could use their CMP settings ("reload page on consent change", first-party cookie deletion) to cover the gap.

packages/experiment-tag/src/types.ts
```diff
@@ -144,6 +163,20 @@ export interface WebExperimentConfig extends ExperimentConfig {
...
    */
   relayUrl?: string;
+  /**
+   * Cookie-consent gating for the web experiment script. When
+   * `consentRequired` is true, the script does not start (no storage access,
+   * evaluation, variant application, tracking, or relay) until the status is
+   * 'granted'. Update status at runtime with
+   * `window.webExperiment.setConsentStatus(status)`.
+   *
+   * v0: `pending`/`denied` defer the start; `granted` starts the client fresh
```
stephen-choi-amplitude: `v0` references. 
1. drop the roadmap promise from public API docs
2. should describe current behavior rather than milestone label

packages/experiment-tag/src/index.ts
```diff
@@ -13,6 +19,52 @@ const eventBuffer: Array<{
...
+// window, so the buffer is left to replay like the normal (non-consent) path.
+const releaseGate = (
+  apiKey: string,
+  initConfigs: InitConfigs,
+  config: WebExperimentConfig,
+): void => {
+  const hadDeferral = consentGate.deferredStart !== null;
+  consentGate.deferredStart = null;
+  if (hadDeferral) {
+    eventBuffer.length = 0;
```
stephen-choi-amplitude: Should the buffer be reset on denied as well?

Also, eventBuffer can grow infinitely in pending state. Any concerns with it?

packages/experiment-tag/src/index.ts
```diff
@@ -74,6 +176,12 @@ const startClient = (
     initConfigs,
     config,
   );
+  // getInstance replaces the stub on globalScope.webExperiment with the real
+  // client. Keep the consent entry point available so a CMP callback that fires
+  // after start still resolves (idempotent grant).
+  (
+    client as unknown as { setConsentStatus?: typeof setConsentStatus }
```
stephen-choi-amplitude: `setConsentStatus` should be typed in `WebExperimentClient` interface and casting shouldn't be required

packages/experiment-tag/src/index.ts
```diff
@@ -23,13 +75,63 @@ export const initialize = (
...
+  // An already-pending deferral keeps the gate closed even if this call resolves
+  // consentRequired=false, so a later initialize can't bypass a start that a
+  // prior one parked on consent.
+  const consent = resolveConsentOptions(config, globalScope);
+  const gated = consent.consentRequired || consentGate.deferredStart !== null;
+  if (gated) {
+    // A runtime status (setConsentStatus) wins over the declarative config.
+    // 'pending' and 'denied' both defer; a later 'granted' — including a
+    // 'denied → granted' re-opt-in — starts the client.
+    const status = consentGate.status ?? consent.consentStatus ?? 'pending';
```
stephen-choi-amplitude: might be worth adding a guardrail for typos

packages/experiment-tag/src/types.ts
tyiuhc: Done in 1bbbd0f — removed the v0/milestone labels and the future-roadmap paragraph. The consentOptions docs now describe only current behavior: pending/denied defer the start, granted starts the client (including denied → granted re-opt-in), pre-grant events are not replayed, and post-start denial is reload-to-reset.

packages/experiment-tag/src/index.ts
tyiuhc: Addressed both in 1bbbd0f. While a start is consent-deferred we no longer buffer at all — releaseGate already discards pre-grant events on grant (consent is not retroactive; replaying them would evaluate triggers against behavior from the no-consent window), so holding them was pure waste and the pending-state growth issue disappears with it. A denied (or pending) status while deferred also clears anything already buffered. The remaining startup-race buffer (non-consent path) is now capped at 500 events, drop-oldest, matching the RTBT event-storage cap.

packages/experiment-tag/src/index.ts
tyiuhc: Done in 1bbbd0f — setConsentStatus is now declared on the WebExperimentClient interface and implemented by DefaultWebExperimentClient (delegating to the module-level function), so the cast in startClient is gone.

packages/experiment-tag/src/index.ts
tyiuhc: Added in 1bbbd0f — statuses are validated via parseConsentStatus. A bad runtime call (e.g. setConsentStatus("grantd")) is a silent no-op; a typo in the declarative consentOptions.consentStatus config logs a warning and is treated as pending, so it fails closed instead of starting the client ungated.

packages/experiment-tag/src/index.ts
tyiuhc: Good catch - there will be a followup to update dev docs

packages/experiment-tag/test/consent/consent-gate.test.ts
```diff
@@ -0,0 +1,356 @@
...
+    };
+    init({
+      consentOptions: { consentRequired: true, consentStatus: 'granted' },
+    });
+    expect(getInstance).not.toHaveBeenCalled();
+    expect(warn).toHaveBeenCalled();
+    warn.mockRestore();
+  });
+
+  test('setConsentStatus with an unknown status is a no-op', () => {
```
stephen-choi-amplitude: Why no-op? Shouldn't this also have the same warning as in initialze?

packages/experiment-tag/src/index.ts
```diff
@@ -23,13 +102,74 @@ export const initialize = (
...
+      setConsentStatus,
+    };
+  }
+
+  // A start already happened (any path — consent-gated or not): don't relaunch.
+  // A second initialize() would otherwise re-fetch preview configs, re-run
+  // start(), or re-open the consent gate against an already-running client.
+  if (consentGate.started) {
+    return;
+  }
```
stephen-choi-amplitude, line 129: This makes a second initialize() a no-op even for non-consent-gated deployments, which is a behavioral change. Is this expected? I think this will affect the chrome extension (https://github.com/amplitude/amplitude-instrumentation-explorer/blob/794e81b33df4a0890519497aa73da14e64252946/app/components/WebExpInject.tsx#L181-L207)

tyiuhc: Good catch that this is a behavioral change — it's intentional.

`consentGate.started` is set on every launch (consent or not) so a later `initialize()` can't re-open the gate against an already-running client. Without that, a non-consent start followed by `initialize({ consentRequired: true, pending })` could stash a deferral and a subsequent grant would launch twice (covered by the "non-consent launch blocks a later consent-gated init" test).

On the chrome extension path specifically: I don't think this breaks the intended flow.

1. The extension already skips when `webExperiment` exists and `isRunning` is true ([WebExpInject.tsx](https://github.com/amplitude/amplitude-instrumentation-explorer/blob/794e81b33df4a0890519497aa73da14e64252946/app/components/WebExpInject.tsx#L181-L207)). The happy path is first-init only (no site SDK → inject → `initialize` with `injectedByExtension`).
2. Even before this change, a second `initialize("chrome-extension", …)` couldn't take over an existing client — `getInstance` returns the singleton and ignores the new `apiKey`/`initConfigs`, and `start()` no-ops once `isRunning` is true.
3. The only overlap window is mid-start (`started === true`, `isRunning === false`). Previously that could re-enter the preview-fetch path and race a concurrent `start()`; the early-return is safer there. The extension still sets `webExperiment.injectedByExtension` after the call.

So the new contract is: `initialize()` is idempotent after the first launch. Happy to tighten docs on that if useful.

packages/experiment-tag/test/consent/consent-gate.test.ts
tyiuhc: Fair — made it symmetric in [`3c85e2d`](https://github.com/amplitude/experiment-js-client/commit/3c85e2d21410cddea31d7efa1bbd8fa7884aa9dd). `setConsentStatus` now warns on an unrecognized status instead of returning silently; the CMP callers are untyped JS so a typo like `grantd` was invisible and left the gate stuck on pending with no signal.

One deliberate difference from the config path: config falls back to `pending` (fail closed, since nothing valid was ever provided), while a bad runtime call keeps the last known status rather than downgrading. Otherwise a stray typo after a real grant would silently revoke consent the CMP already gave. Added a test for that case.

kwliou (approved):

PR #355, opened 2026-07-29
feat(experiment-browser): support injectable in-memory cache storage for web experiment

tyiuhc:
## Summary

Lets the web experiment client take an injectable cache storage instead of always writing its `amp-exp-*` variant/flag/options caches to `sessionStorage`. This is the one `experiment-browser` piece needed for consent v1 (no-flicker in-memory `pending`), where experiments must evaluate before the end user has agreed to any storage.

**Ships dark.** Nothing injects it yet — the default path is unchanged, so this is a no-op for every current deployment. `experiment-tag` starts passing `MemoryStorage` when the consent-gated in-memory mode lands.

Tracked by [WEB-173](https://linear.app/amplitude/issue/WEB-173). Design: [Cookie Consent Control — Tech Spec](https://linear.app/amplitude/document/cookie-consent-control-tech-spec-1e85897e46f7) §6b.

## What changed

- **`storage/memory-storage.ts`** (new) — `MemoryStorage`, a `Map`-backed `Storage` mirroring `local-storage.ts` / `session-storage.ts`. Exported from `src/index.ts` so `experiment-tag` can construct it.
- **`experimentClient.ts`** — the web-experiment branch of the storage selection now reads an internal `internalCacheStorage` off the config and **always falls back to `SessionStorage`** when it's absent. Non-web clients are untouched and keep `LocalStorage`.
- **`test/memoryStorage.test.ts`** (new) — 5 tests.

## Risk

- **Fallback is unconditional.** The POC this derives from could leave `storage` undefined when nothing was injected (flagged by Bugbot); here the `?? new SessionStorage()` fallback is covered by a dedicated test, so the default path cannot regress to undefined storage.
- **Internal, not public config.** `internalCacheStorage` is read via index access like the existing `internalInstanceNameSuffix`, so it stays off the documented `ExperimentConfig` surface.
- **Storage is chosen once, at construction.** After a mid-session consent grant the cache stays in memory for the page lifetime and reverts to `sessionStorage` on the next load. Safe because it's a cache — it repopulates on the next fetch, and no flush is needed.
- **Non-web unaffected.** A test asserts an injected storage is ignored (and `LocalStorage` still used) when the client isn't the web instance.

## Test plan

- [x] `yarn test` in `experiment-browser` — 212 passing, incl. 5 new
- [x] `yarn lint` — clean
- [x] `yarn build` — emits `dist/types/src/storage/memory-storage.d.ts`; `MemoryStorage` present in all three bundles (needed before `experiment-tag` type-checks against it)

### Checklist

[WEB-173]: https://amplitude.atlassian.net/browse/WEB-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/experiment-browser/test/memoryStorage.test.ts
+++ b/packages/experiment-browser/test/memoryStorage.test.ts
@@ -1,0 +1,93 @@
+import { ExperimentClient } from '../src/experimentClient';
+import { MemoryStorage } from '../src/storage/memory-storage';
+import { Storage } from '../src/types/storage';
+
+const API_KEY = 'client-DvWljIjiiuqLbyjqdvBaLFfEBrAvGuA3';
+
+class SpyStorage implements Storage {
+  gets: string[] = [];
+  puts: Record<string, string> = {};
+  get(key: string): string {
+    this.gets.push(key);
+    return undefined;
+  }
+  put(key: string, value: string): void {
+    this.puts[key] = value;
+  }
+  delete(key: string): void {
+    delete this.puts[key];
+  }
+}
+
+describe('MemoryStorage', () => {
+  test('get returns what was put, delete removes', () => {
+    const storage = new MemoryStorage();
+    expect(storage.get('k')).toBeUndefined();
+    storage.put('k', 'v');
+    expect(storage.get('k')).toEqual('v');
+    storage.delete('k');
+    expect(storage.get('k')).toBeUndefined();
+  });
+
+  test('instances are isolated', () => {
+    const a = new MemoryStorage();
+    const b = new MemoryStorage();
+    a.put('k', 'v');
+    expect(b.get('k')).toBeUndefined();
+  });
+});
+
+describe('web experiment cache storage selection', () => {
+  let sessionGetItem: jest.SpyInstance;
+  let localGetItem: jest.SpyInstance;
+
+  beforeEach(() => {
+    sessionGetItem = jest.spyOn(
+      Object.getPrototypeOf(window.sessionStorage),
+      'getItem',
+    );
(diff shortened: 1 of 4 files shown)


tyiuhc:
bugbot run

packages/experiment-browser/test/memoryStorage.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  beforeEach(() => {
+    sessionGetItem = jest.spyOn(
+      Object.getPrototypeOf(window.sessionStorage),
+      'getItem',
+    );
+    localGetItem = jest.spyOn(
+      Object.getPrototypeOf(window.localStorage),
+      'getItem',
+    );
```
ansonchen2: nit: not sure if this matters, but I think `sessionStorage` and `localStorage` share one prototype, and `jest.spyOn` on an already mocked property returns the existing mock rather than a new one, meaning `localGetItem` actually registers on the session spy similar to `sessionGetItem`

ansonchen2 (approved):

packages/experiment-browser/test/memoryStorage.test.ts
tyiuhc: Good catch — fixed in 872e959. jsdom also ignores instance-level `getItem` assignment/`spyOn`, so we spy `Storage.prototype` once and route by `this` to keep session vs local distinguishable. Each selection test now also asserts the unused store was not touched.

PR #358, opened 2026-07-29
feat(experiment-tag): hold storage in memory while consent is pending

tyiuhc:
Part of [WEB-171](https://linear.app/amplitude/issue/WEB-171/gate-storage-surfaces-on-consent-localstorage-cookies-rtbt-session) — every storage surface the tag touches, gated on consent. Impression buffering (WEB-172's other half) is stacked on top of this one.

**Stacked on #357**, which supplies the denial sweep this PR's refusal path is designed to coexist with. Review that first; this is the delta on top of it.

## The rule this implements

While the visitor has yet to answer the banner, nothing the tag would persist reaches the device — it is held in memory instead. A grant hands it over verbatim. A refusal drops it.

That needs two predicates, not one:

- `isConsentPending()` — undecided **and** gating is on. The data is worth holding, because consent may still arrive.
- `isConsentWithheld()` — gating is on and consent is not in hand, whether undecided or refused. This is the condition for keeping data off the device.

Refusal has to suppress writes and not merely trigger #357's sweep. A visitor who revokes mid-visit leaves a client already running, and erasing its data while it carries on writing puts the data straight back.

Both predicates are false whenever `consentRequired` was never set, so every gate here is inert for the overwhelming majority of pages. They read `consentGate.required`, a latch set in `initialize()` before the client is constructed: the manager starts at `pending` for everyone, so the status alone cannot tell "still deciding" from "feature off", and a second `initialize()` resolving `consentRequired: false` must not reopen storage a first one closed.

## Reads are gated alongside writes

ePrivacy covers *access* to data already on the device, not only storing it. So a pending visitor reads what this page buffered and nothing else — never what an earlier consented visit left behind. After a refusal the buffer is empty, so every read comes back absent.

That is also what stops re-identification: the RTBT session cookie, the cross-subdomain identity cookie, and the behavioral-events store all read as blank until consent arrives.

## The four surfaces

| Surface | While pending |
|---|---|
| `util/storage.ts` — identity, `DEFAULT_USER_PROVIDER`, redirect sessionStorage, `EXP_MKTG_*`, preview | Buffered per `storageType:key`. Entries hold the serialized form, so a buffered read parses a fresh copy exactly as a real read would and a flush is a verbatim handover. |
| `SyncJsonCookie` — RTBT session | Degrades to the memory tier it already has. On grant it re-writes **the same value**, which is what keeps the session id the visitor had while the banner was up instead of rotating it at the moment of grant. |
| `ConsentAwareCookieStorage` — identity / redirect / marketing cookies that go through analytics-core | Buffered in the wrapper. Reads and writes join the in-flight grant flush, so nothing observes the store half-populated. |
| RTBT event store | Cache starts empty and the debounced localStorage write is suppressed. `hasPendingWrites` is deliberately left set so a grant persists everything gathered while the banner was up. |

## `remove` is deliberately not symmetric with `set`

Removal is held back only while **pending**, and only as far as the buffer — issuing the expiry or delete is itself a write to the device. Refusal falls through to real storage on purpose, because that is the path #357's cleanup uses to erase what a previously consented visit left behind. The same asymmetry appears in `SyncJsonCookie.clear` and `ConsentAwareCookieStorage.remove`.

`SyncJsonCookie.clear` also skips domain resolution while pending, since the resolver probes by *setting* a throwaway cookie.

## Two keys stay exempt

`amp-preview-mode` and `visual-editor-state` remain readable and writable in every state. Both hold Amplitude's own tooling state, entered deliberately by whoever is building the experiment via a `PREVIEW` / `VISUAL_EDITOR` URL parameter — they say nothing about the visitor and are not what a consent banner is asking about. Gating them would leave the editor silently failing to open on any site with consent gating switched on, because that state has to survive the redirect that strips those parameters from the URL.

They move to a new `util/storage-keys.ts` so the gate can name them without importing the modules that call it. `experiment.ts` and `util/messenger.ts` re-export their previous names, so no call site changed.

## Subscription discipline

Pending resolves exactly once, so every flush subscription unsubscribes on the first transition either way. That is also the mechanism that stops a value gathered before a refusal from being written out later, should the visitor return to the banner and opt in — each suite pins that case.

In `util/storage.ts` the listener is tracked by manager **instance** rather than a boolean, so `consentGate.reset()` (tests only) re-arms on the next gated call instead of leaving the buffer wired to a manager nothing transitions any more.

## Ships partly inert, on purpose

`ConsentAwareCookieStorage` and `createCookieStorage` have no call sites yet, and `resolveCrossSubdomainObject` is loosened to accept the interface so they can be dropped in. Routing the three real `CookieStorage` sites through them — along with `initializeConsent` ordering, relay deferral, and the redirect URL transport — is [WEB-209](https://linear.app/amplitude/issue/WEB-209/wire-consent-into-indexts-and-experimentts-init-relay-redirect-cookie). Split that way so this PR reads as "the gates" and that one as "the wiring".

A consequence worth stating: on its own this PR does not yet make a pending page storage-free end to end. It makes every gated helper correct; WEB-209 is what puts the remaining cookie sites on those helpers.

## Test plan

- [x] `consent-storage.test.ts` — feature-off pass-through; pending buffering keyed per store, no read of an earlier consented session, fresh copy per read, last-write-wins, remove drops the buffered entry and leaves the persisted key alone; grant flushes each entry to the store it was destined for and does not flush a key removed while pending; denial discards, keeps dropping later writes, reads absent, still removes from real storage, and never replays after a later grant; all four exemption behaviors; `reset` re-arms against the replacement manager
- [x] `consent-sync-cookie.test.ts` — pending serves reads from memory, ignores a cookie from a consented session, issues no expiry on `clear`, and does not resolve the domain (which would probe by writing); grant promotes the deferred value **keeping the same id** and flushes the latest of several writes; denial never promotes, reads absent, still deletes on `clear`, and does not flush after a later grant
- [x] `consent-cookie-storage.test.ts` — pending buffers and serves reads; grant hands over to the delegate, never exposes a half-flushed store to a racing read, and survives a delegate whose write is blocked; denial discards and still removes
- [x] `consent-event-storage.test.ts` — pending keeps events in memory, does not load a consented visit's events, and does not rewrite the store to backfill uuids on load; grant persists what accumulated; denial persists nothing, including on the teardown flush
- [x] 404 tests / 25 suites green (up from 337 / 21 on #357 — 4 new suites, 67 new tests), with every pre-existing test unchanged since consent is off by default
- [x] `yarn lint` 0 errors (69 warnings, all pre-existing repo-wide); `tsc --noEmit` clean

[WEB-171]: https://amplitude.atlassian.net/browse/WEB-171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-209]: https://amplitude.atlassian.net/browse/WEB-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
+++ b/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
@@ -1,9 +1,9 @@
 import { getGlobalScope } from '@amplitude/experiment-core';
 
 import {
-  consentGate,
   isConsentPending,
   isConsentWithheld,
+  onConsentDecision,
 } from '../consent/consent-gate';
 
 import { RelayClient } from './relay-client';
@@ -425,20 +425,14 @@ export class EventStorageManager {
 
   /**
    * Persists the events gathered while consent was pending, once it is granted.
-   *
-   * Pending resolves only one way or the other, so the subscription is spent on
-   * the first transition either way. Dropping it on refusal is what stops events
-   * collected before that refusal from reaching the device later, should the
-   * visitor return to the banner and opt in.
    */
   private armConsentFlush(): void {
     if (this.consentFlushArmed) {
       return;
     }
     this.consentFlushArmed = true;
-    const unsubscribe = consentGate.manager.onChange((status) => {
-      unsubscribe();
-      if (status === 'granted') {
+    onConsentDecision((granted) => {
+      if (granted) {
         this.flushToLocalStorage();
       }
     });
--- a/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
+++ b/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
@@ -1,5 +1,10 @@
 import { getGlobalScope } from '@amplitude/experiment-core';
 
+import {
+  isConsentPending,
+  isConsentWithheld,
+  onConsentDecision,
+} from '../consent/consent-gate';
 import { getStorageItem, setStorageItem } from '../util/storage';
 
 import { RelayClient } from './relay-client';
@@ -74,6 +79,7 @@ export class EventStorageManager {
   private persistedEvents?: Set<string>; // Optional set of event types to persist
   private storageKey: string;
   private relayClient: RelayClient | null = null;
+  private consentListenerArmed = false;
 
   constructor(
     apiKey: string,
@@ -118,6 +124,10 @@ export class EventStorageManager {
     };
 
     this.memoryCache.events.push(event);
+
+    if (isConsentPending()) {
+      this.armConsentListener();
+    }
 
     // Apply FIFO limit. This bounds local memory only; the relay keeps its own
     // copy (the event was already dual-written below), so no relay
@@ -387,6 +397,18 @@ export class EventStorageManager {
       return; // No changes to persist
     }
 
+    if (isConsentWithheld()) {
+      if (isConsentPending()) {
+        setStorageItem('localStorage', this.storageKey, this.memoryCache);
+      }
+      this.hasPendingWrites = false;
+      if (this.debouncedWriteTimeout) {
+        clearTimeout(this.debouncedWriteTimeout);
+        this.debouncedWriteTimeout = null;
+      }
+      return;
+    }
+
     setStorageItem('localStorage', this.storageKey, this.memoryCache);
     this.hasPendingWrites = false;
 
@@ -395,6 +417,28 @@ export class EventStorageManager {
       clearTimeout(this.debouncedWriteTimeout);
       this.debouncedWriteTimeout = null;
     }
+  }
+
+  /**
+   * Pending-window RTBT events must not reach the device on a later re-grant.
+   * Only armed while consent is still pending at first write.
+   */
+  private armConsentListener(): void {
+    if (this.consentListenerArmed) {
+      return;
+    }
+    this.consentListenerArmed = true;
+    onConsentDecision((granted) => {
+      if (granted) {
+        return;
+      }
+      this.memoryCache = { events: [], nextId: 1 };
+      this.hasPendingWrites = false;
+      if (this.debouncedWriteTimeout) {
+        clearTimeout(this.debouncedWriteTimeout);
+        this.debouncedWriteTimeout = null;
+      }
+    });
   }
 
   /**
--- a/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
+++ b/packages/experiment-tag/src/behavioral-targeting/event-storage.ts
@@ -1,5 +1,11 @@
 import { getGlobalScope } from '@amplitude/experiment-core';
 
+import {
+  consentGate,
+  isConsentPending,
+  isConsentWithheld,
+} from '../consent/consent-gate';
+
 import { RelayClient } from './relay-client';
 import { RelayEventStorage } from './relay-protocol';
 import { SessionManager } from './session-manager';
@@ -72,6 +78,7 @@ export class EventStorageManager {
   private persistedEvents?: Set<string>; // Optional set of event types to persist
   private storageKey: string;
   private relayClient: RelayClient | null = null;
+  private consentFlushArmed = false;
 
   constructor(
     apiKey: string,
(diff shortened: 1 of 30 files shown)


packages/experiment-tag/src/behavioral-targeting/event-storage.ts
```diff
@@ -380,23 +376,24 @@ export class EventStorageManager {
...
-      this.hasPendingWrites = false;
+    setStorageItem('localStorage', this.storageKey, this.memoryCache);
+    this.hasPendingWrites = false;
 
-      // Clear debounce timeout since we just flushed
-      if (this.debouncedWriteTimeout) {
-        clearTimeout(this.debouncedWriteTimeout);
-        this.debouncedWriteTimeout = null;
-      }
-    } catch (e) {
```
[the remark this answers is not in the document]
tyiuhc: Pre-existing behavior, not introduced here — `flushToLocalStorage` has cleared `hasPendingWrites` unconditionally after the write since EventStorage landed in #262; this PR only reroutes the write through the consent-gated helper without changing that. A real fix needs `setStorageItem` to report success (it currently swallows quota/availability errors), which touches every caller of the storage util — worth a follow-up outside the consent stack rather than a rider on this PR. Note the consent gate doesn't widen the window: while consent is pending the flush isn't attempted at all, and the buffered-flush path on grant goes through the same pre-existing write.

packages/experiment-tag/src/consent/consent-gate.ts
```diff
@@ -43,6 +53,7 @@ interface ConsentGate {
  */
 export const consentGate: ConsentGate = {
```
kwliou, line 55: I don't know how many pieces of state there are when the SDK runs but if there are multiple cases similar to this, it can be frustrating to debug. When I'm debugging, I would prefer to inspect a single root object like `window.webExperiment` to have a view of the entire current state.

tyiuhc: Agreed — added in cb48e42. The tag already has a single debug root: `window.webExperiment.getDebugState()` (enable with `window.__AMP_DEBUG = true` or the `amp-ve-debug` localStorage key). It now includes a `consent` section: `status`, `required`, `started`, and `startDeferred`. The state stays module-scoped because it must exist before the client does, but in the state you'd debug (pending) the client is running, so the snapshot is reachable there. Impression-buffer counts join the same section in #359.

packages/experiment-tag/src/index.ts
kwliou, line 133: Could you clarify for me if this is by design? Do you know how the analytics SDK handles this? Is it because those events get dropped pre-consent as well that we should too so that RTBT stays accurate (event never happened)? If the analytics event also gets "buffered" as well then it seems like we should too.

packages/experiment-tag/src/util/storage.ts
```diff
@@ -1,7 +1,67 @@
...
+const pendingWrites = new Map<
+  string,
+  { storageType: StorageType; key: string; json: string }
+>();
+
+/**
+ * The manager this module's flush/drop listener is attached to. Tracking the
+ * instance rather than a boolean means a replaced manager (only `reset` does
+ * that) re-arms on the next gated call instead of leaving the buffer wired to a
+ * manager nothing transitions any more.
```
kwliou: "leaving the buffer wired to a manager nothing transitions any more."  I don't understand this phrase.

packages/experiment-tag/src/index.ts
tyiuhc: By design. The analytics SDK has no consent gate of its own — in a standard CMP setup analytics doesn't start until consent, so pre-grant events never exist and "event never happened" is the consistent posture. If a customer runs analytics ungated, those events reach Amplitude through their pipeline regardless — this drop only decides whether RTBT sees them. And by #361 the window is narrow: under `pending` the client runs and events flow into the consent-gated in-memory event store (evaluated live, persisted on grant), so a deferral — and this drop — only happens for denied-at-load, where nothing may be retained. Buffering-for-replay also fits impressions but not RTBT events: an impression replayed with its stamped `time` is still a correct fact; RTBT events are react-now trigger signals, and a replayed batch at grant time would fire time-sensitive triggers late and all at once.

packages/experiment-tag/src/util/storage.ts
tyiuhc: Reworded in 97b1ba5 — it was trying to say: the test-only `consentGate.reset()` replaces the manager with a fresh instance, which strands any listener subscribed to the old one (its status never changes again, so the flush/drop would never fire). Comparing instances instead of tracking an "armed" boolean makes the next gated call notice the swap and re-subscribe to the live manager.

kwliou (approved):

PR #359, opened 2026-07-29
feat(experiment-tag): buffer impressions until consent is granted

tyiuhc:
Closes the second half of [WEB-172](https://linear.app/amplitude/issue/WEB-172/rejection-cleanup-impression-buffering). The rejection-cleanup half is #357; this is impression buffering, which needed WEB-171's gate primitives to exist first.

**Stacked on #358.** Review that first — this is one new file plus a five-line hook.

## What it does

A visitor who leaves the consent banner up still gets variants applied, which means impressions still fire. Those impressions are the last thing in the tag that would reach the device without consent, and they take an unusual route there: the SDK's `PersistentTrackingQueue` parks any unsent event in `EXP_unsent_*` localStorage.

So the buffer sits **upstream** of that queue, wrapping the integration's `track` before `addPlugin` binds it. Every deferred impression reports itself as tracked, so the queue lets go of it and the payload never reaches localStorage — this buffer becomes the only thing keeping it. On grant they replay; on refusal they are dropped.

That `return true` is the load-bearing decision, and it reads as a lie about a refused impression. The alternative is worse: returning `false` hands the event straight to the persistent queue, which is exactly the localStorage write consent forbids. The plan text said to return `false` on refusal; this implements `true` for that reason.

## Replay carries the original timestamp

The analytics SDK timestamps an event as it receives it, which for a replay is the moment of the grant — potentially minutes after the variant was actually shown. Each buffered impression gets `eventProperties.time` stamped when it occurs, and an explicit `time` already on the event is left alone. Without this, every consent-deferred impression is misattributed to the grant.

The stamp is applied to a copy; the caller's event object is not mutated.

## Replay is ordered and retried

Impressions replay in the order they occurred, stopping at the first one the integration declines — the analytics SDK it forwards to may not have a receiver attached yet at the moment of grant. The rest stay buffered and a 1s poller retries until the buffer drains, mirroring how the SDK's own queue behaves. A throwing integration is treated as a decline rather than losing the event.

The buffer is capped at 512 events, matching `PersistentTrackingQueue`'s own cap, and drops oldest-first.

## Idempotent wrap

The integration object is shared with the customer — they may have supplied it — and two copies of the tag can land on one page. `wrapIntegrationTrack` marks the integration with a string-keyed flag (not a symbol, so separate bundles recognize each other's wrapper) and returns early if it is already gated.

## Ordering constraint

The wrap must precede `addPlugin`: the integration manager binds `track` as it takes the plugin on, so a wrap applied afterwards would never be seen. That is why the call sits directly after `experimentIntegration.type = 'integration'` rather than anywhere later in `start()`.

## Test plan

- [x] `consent-impression-buffer.test.ts` — feature-off and granted-from-the-start track straight through, and the integration's own return value is relayed so the SDK can still retry a refusal; double-wrap is a no-op
- [x] Pending — impression withheld but reported as tracked, occurrence time stamped, an explicit `time` preserved, caller's event object not mutated
- [x] Grant — deferred impressions sent in the order they occurred, direct pass-through afterwards, retry for an integration that is not yet ready, poller stops once the buffer drains, a throwing integration is retried rather than dropped
- [x] Denial — impressions deferred while pending are dropped, are not sent if consent arrives later, and later impressions are discarded rather than tracked
- [x] 420 tests / 26 suites green (up from 404 / 25 on #358 — 1 new suite, 16 new tests)
- [x] `yarn lint` 0 errors (69 warnings, all pre-existing repo-wide); `tsc --noEmit` clean

[WEB-172]: https://amplitude.atlassian.net/browse/WEB-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/experiment-tag/src/consent/consent-impression-buffer.ts
+++ b/packages/experiment-tag/src/consent/consent-impression-buffer.ts
@@ -1,0 +1,149 @@
+import { getClearInterval, getSetInterval } from '@amplitude/experiment-core';
+import type {
+  ExperimentEvent,
+  IntegrationPlugin,
+} from '@amplitude/experiment-js-client';
+
+import {
+  isConsentPending,
+  isConsentWithheld,
+  onConsentDecision,
+} from './consent-gate';
+
+/** Matches the cap on the SDK's own PersistentTrackingQueue. */
+const MAX_BUFFERED_EVENTS = 512;
+const RETRY_INTERVAL_MS = 1000;
+
+/**
+ * Marks an integration whose `track` is already gated. A string key rather than
+ * a symbol so that two copies of the tag on one page recognize each other's
+ * wrapper and the integration is not gated twice.
+ */
+const WRAPPED_KEY = '__ampConsentGatedTrack';
+
+type Tracker = (event: ExperimentEvent) => boolean;
+
+/**
+ * Holds impressions in memory while the visitor has yet to answer the consent
+ * banner, and sends them once consent is granted.
+ *
+ * This sits upstream of the SDK's PersistentTrackingQueue, which would otherwise
+ * park an unsent impression in localStorage. Every deferred impression therefore
+ * reports itself as tracked, so the queue lets go of it and the payload never
+ * reaches the device — this buffer becomes the only thing keeping it.
+ */
+class ImpressionBuffer {
+  private readonly buffered: ExperimentEvent[] = [];
+  private flushArmed = false;
+  private poller: ReturnType<typeof setInterval> | undefined;
+
+  constructor(private readonly realTrack: Tracker) {}
+
+  track(event: ExperimentEvent): boolean {
+    if (!isConsentWithheld()) {
+      return this.realTrack(event);
+    }
+    if (isConsentPending()) {
+      this.buffer(event);
+      this.armFlush();
+    }
+    // A refusal discards the impression: it is the one outcome where the event
+    // must never be sent, so there is nothing to hold on to.
+    return true;
+  }
+
+  private buffer(event: ExperimentEvent): void {
+    if (this.buffered.length >= MAX_BUFFERED_EVENTS) {
+      this.buffered.shift();
+    }
+    this.buffered.push(stampTime(event));
+  }
+
+  private armFlush(): void {
+    if (this.flushArmed) {
+      return;
+    }
+    this.flushArmed = true;
+    onConsentDecision((granted) => {
+      if (granted) {
+        this.flush();
+      } else {
+        this.buffered.length = 0;
+      }
+    });
+  }
+
+  /**
+   * Replays the deferred impressions in the order they occurred, stopping at the
+   * first one the integration will not take — the analytics SDK it forwards to
+   * may not have a receiver attached yet. The rest stay buffered and a poller
+   * retries, mirroring how the SDK's own queue drains.
+   */
+  private flush(): void {
+    let sent = 0;
+    for (; sent < this.buffered.length; sent++) {
+      try {
+        if (!this.realTrack(this.buffered[sent])) {
+          break;
+        }
+      } catch {
+        break;
+      }
+    }
+    this.buffered.splice(0, sent);
+    if (this.buffered.length > 0) {
+      this.startRetry();
+    } else {
+      this.stopRetry();
+    }
+  }
+
+  private startRetry(): void {
+    if (this.poller !== undefined) {
+      return;
+    }
+    const setInterval = getSetInterval();
+    if (!setInterval) {
+      return;
+    }
+    this.poller = setInterval(() => {
+      this.flush();
+    }, RETRY_INTERVAL_MS);
+  }
+
+  private stopRetry(): void {
+    if (this.poller === undefined) {
+      return;
+    }
+    getClearInterval()?.(this.poller);
+    this.poller = undefined;
+  }
+}
+
+/**
+ * Records when the impression happened. The analytics SDK timestamps an event
+ * as it receives it, which for a replay is the moment of the grant — long after
+ * the variant was shown for a visitor who leaves the banner up.
+ */
+function stampTime(event: ExperimentEvent): ExperimentEvent {
+  const properties = event.eventProperties;
+  if (!properties || properties.time !== undefined) {
+    return event;
+  }
+  return { ...event, eventProperties: { ...properties, time: Date.now() } };
+}
+
+/**
+ * Routes an integration's impressions through a consent buffer, in place. The
+ * integration is shared with the customer (they may have supplied it), so the
+ * wrapper is applied at most once.
+ */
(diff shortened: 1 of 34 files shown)


packages/experiment-tag/src/consent/consent-impression-buffer.ts
```diff
@@ -0,0 +1,151 @@
...
+const RETRY_INTERVAL_MS = 1000;
+
+/**
+ * Marks an integration whose `track` is already gated. A string key rather than
+ * a symbol so that two copies of the tag on one page recognize each other's
+ * wrapper and the integration is not gated twice.
+ */
+const WRAPPED_KEY = '__ampConsentGatedTrack';
+
+type Tracker = (event: ExperimentEvent) => boolean;
```
kwliou, line 34: It's not clear what this `Tracker` type represents e.g. what is the return value for? Could you add a comment here or to the constructor below?

packages/experiment-tag/src/consent/consent-impression-buffer.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+/**
+ * Records when the impression happened. The analytics SDK timestamps an event
+ * as it receives it, which for a replay is the moment of the grant — long after
+ * the variant was shown for a visitor who leaves the banner up.
+ */
+function stampTime(event: ExperimentEvent): ExperimentEvent {
+  if (event.eventProperties?.time !== undefined) {
```
kwliou, line 166: nit: `time` as a label isn't really specific.

packages/experiment-tag/src/consent/consent-impression-buffer.ts
kwliou, line 34: Could you comment here or the constructor below. It's not clear what this represents and is used for e.g. what is the return value supposed to be?

packages/experiment-tag/src/consent/consent-impression-buffer.ts
```diff
@@ -0,0 +1,151 @@
...
+ * Routes an integration's impressions through a consent buffer, in place. The
+ * integration is shared with the customer (they may have supplied it), so the
+ * wrapper is applied at most once.
+ */
+export function wrapIntegrationTrack(integration: IntegrationPlugin): void {
+  const marked = integration as IntegrationPlugin & Record<string, unknown>;
+  if (marked[WRAPPED_KEY]) {
+    return;
+  }
+  const buffer = new ImpressionBuffer(integration.track.bind(integration));
```
kwliou, line 203: Not blocking but same case in the other PR where I'm worried this won't be straightforward to debug in prod i.e. can I easily view the impression buffer state? I could set a breakpoint if I can find the poller function but the code would be minified. It would be nice if I could view the multiple queues in one place or if it was exposed on the SDK instance somewhere.

packages/experiment-tag/src/consent/consent-impression-buffer.ts
tyiuhc: Documented in a5e6948 — it's the integration's original `track`, and the boolean is `IntegrationPlugin.track`'s contract: whether the downstream integration accepted the event (the analytics SDK may not have a receiver attached yet). `flush` relies on it — `false` or a throw stops the replay, keeps the rest buffered, and hands off to the retry poller.

tyiuhc: Documented in a5e6948 (see thread above) — the return value is what stops a flush when the analytics SDK has no receiver yet.

packages/experiment-tag/src/consent/consent-impression-buffer.ts
tyiuhc: The key is a contract rather than a label: the analytics SDK's event-bridge receiver destructures `time` out of `eventProperties` and promotes it to the canonical event timestamp (`setEventReceiver` in Amplitude-TypeScript's `browser-client.ts`), and the client already stamps `exposure.time` on web experiment exposures — this is the fallback for events that arrive without one. Renaming would silently break replay timestamps. Added a comment citing this in a5e6948.

packages/experiment-tag/src/consent/consent-impression-buffer.ts
tyiuhc: Added in a5e6948 — `window.webExperiment.getDebugState().consent.impressionBuffers` now reports one entry per wrapped integration: `buffered` count, `flushArmed`, and whether the retry poller is running. Same debug root as the gate state added in #358.

packages/experiment-tag/src/experiment.ts
```diff
@@ -619,6 +621,11 @@ export class DefaultWebExperimentClient implements WebExperimentClient {
       );
     }
     this.globalScope.experimentIntegration.type = 'integration';
+    if (consentGate.required) {
+      // Must precede addPlugin: the integration manager binds `track` as it
+      // takes the plugin on, so a later wrap would never be seen.
+      wrapIntegrationTrack(this.globalScope.experimentIntegration);
```
kwliou, line 627: Confirming, does putting this here make sense? It looks like start() needs to wait for consent anyways (deferredStart)?

tyiuhc: It does — start() having run doesn't mean consent is settled. The wrapper covers (1) mid-session revocation: a visitor can deny after granting while the client keeps running, and `isConsentWithheld()` is what discards impressions from that moment; and (2) #361, which starts the client under `pending` — this PR introduces the buffer, that one routes pending traffic through it. It sits at this exact point because the integration manager binds `track` as it takes the plugin on; a later wrap would never be seen. `consentGate.required` keeps it a no-op when gating is off.

kwliou (commented):
I've been thinking of ways to consolidate the surface areas of the multiple flushes and buffers. Ideally this would also make it more deterministic and easier to follow. What do you think about refactoring each of the instances of ConsentAwareCookieStorage, SyncCookie, ImpressionBuffer etc. communicating with a single ConsentRuntime/Manager which handles the consent state changes (and pending sink queues etc) rather than each of them reading consent status.

kwliou:
We can discuss larger refactors later on but here's my understanding with AI:
Web storage flushes its map.
Cookie storage flushes another map.
SyncJsonCookie promotes its memory value.
ImpressionBuffer starts replaying.
Denial cleanup erases persisted data.

I see benefits in centralizing and coordinating this transition versus the multiple localized callbacks.

flushPendingState() {
  await this.flushPhase(Identity)
  await this.flushPhase(Session)
  await this.flushPhase(Attribution)
  await this.flushPhase(CachedState)
  await this.drain(Impressions)
}

kwliou (approved):

